PR #8, opened 2020-03-16
Clean up repo by enabling a large number of clippy lints

spacejam:
This is a lot of churn, and I'm happy to break things up to avoid conflicts etc... if that's helpful! 

Summary of changes:
* added docs on all public types
* try to use `const fn` where possible instead of inlined functions
* avoid `inline(always)` and prefer `inline` as often the compiler will do a better job of optimizing
* use `eprintln` for now to avoid silent errors in background threads. A likely patch coming soon will use the `log` crate, and also it's likely that any error here should trigger either a process abort *or* setting a `Connection`-specific error that will feed back into all `Result`-returning methods for user feedback. There should never be a vital background thread that dies without humans being able to react via seeing a crash or receiving errors on their called methods.
* generally, just doing what clippy suggests most of the time, and in a few very rare cases (due to serde skip serialization function signature requirements) mark a clippy override


src/lib.rs
```diff
@@ -90,12 +166,18 @@ where
     }
 }
 
-#[derive(Debug, PartialEq)]
+/// A `ConnectionStatus` describes the current sub-status of a `Connected` connection.
+#[derive(Debug, PartialEq, Clone, Copy)]
```
spacejam: `Copy` implementations generally allow the compiler to make more aggressive optimizations, and are useful on pretty much any type that you don't implement `Drop` on.

src/lib.rs
```diff
@@ -109,18 +191,18 @@ pub(crate) struct Outbound {
 }
 
 impl Outbound {
-    #[inline(always)]
+    #[inline]
```
spacejam: The compiler has more information about when inlining is appropriate. `inline(always)` can actually slow down programs by causing the code section to grow in size, reducing the efficiency of your CPU's instruction cache. LLVM is usually pretty good about reasoning about where that line between instruction prefetcher-friendliness and instruction cache-friendliness is.

src/lib.rs
```diff
@@ -109,18 +191,18 @@ pub(crate) struct Outbound {
...
     fn write_response(&mut self, subj: &str, msgb: &[u8]) -> io::Result<()> {
         write!(self.writer, "PUB {} {}\r\n", subj, msgb.len())?;
-        self.writer.write(msgb)?;
-        self.writer.write(b"\r\n")?;
+        self.writer.write_all(msgb)?;
+        self.writer.write_all(b"\r\n")?;
```
spacejam: This is an actual bug, because `write` may return before completely writing the desired bytes, resulting in short writes. `write_all` loops until the full buffer is written.

src/lib.rs
```diff
@@ -186,7 +271,8 @@ impl Connection<NotConnected> {
     /// # Ok(())
     /// # }
     /// ```
-    pub fn new() -> Connection<NotConnected> {
+    #[allow(clippy::new_without_default)]
+    pub const fn new() -> Connection<NotConnected> {
```
spacejam: `const fn` is even better than `inline(always)` because it gets completely evaluated at compile time.

src/lib.rs
```diff
@@ -209,7 +295,7 @@ impl Connection<NotConnected> {
     /// # }
     /// ```
     pub fn with_token(self, token: &str) -> Connection<Authenticated> {
-        let mut opts = self.options.clone();
```
spacejam: No need to clone the backing options because this is a `self` function that takes ownership over them anyway.

src/lib.rs
```diff
@@ -241,21 +327,23 @@ impl Connection<NotConnected> {
...
-                Some(_) => nats_url.to_string(),
-                None => format!("{}:4222", nats_url),
-            },
+            Err(_) => {
+                if nats_url.find(':').is_some() {
+                    nats_url.to_string()
+                } else {
+                    format!("{}:4222", nats_url)
+                }
+            }
```
spacejam: A rule of thumb for when to use match or an if: use match when you need more than 2 arms, otherwise if/if-let

src/lib.rs
```diff
@@ -334,7 +423,7 @@ impl Connection<Authenticated> {
             thread::yield_now();
             let mut cur_len = wbuf.lock().unwrap().writer.buffer().len();
             if cur_len != start_len {
-                for d in wait.iter() {
+                for d in &wait {
```
spacejam: The idiom is to avoid `.iter()` methods on types that have implemented `IntoIterator` for `&T`. for-loops look for an implementation of the `Iterator` trait, falling back to `IntoIterator` if `Iterator` is not implemented

src/lib.rs
```diff
@@ -577,7 +675,7 @@ impl Subscription {
             for m in r.iter() {
                 if let Err(e) = handler(m) {
                     // TODO(dlc) - Capture for last error?
-                    println!("Error in callback! {:?}", e);
+                    eprintln!("Error in callback! {:?}", e);
```
spacejam: These calls to `eprintln` (same as println but writes to stderr) should probably be replaced by using the `log` crate and using the `error!` macro, or going further and using structured logging.

src/lib.rs
```diff
@@ -686,10 +784,7 @@ pub struct SubscriptionDeadlineIterator {
 impl Iterator for SubscriptionDeadlineIterator {
     type Item = Message;
     fn next(&mut self) -> Option<Self::Item> {
-        match self.r.recv_timeout(self.to) {
-            Ok(m) => Some(m),
-            _ => None,
-        }
+        self.r.recv_timeout(self.to).ok()
```
spacejam: This is a handy built-in method for "downgrading" a Result to an Option

src/lib.rs
```diff
@@ -966,7 +1061,16 @@ impl Connection<Connected> {
         match parser::parse_control_op(reader)? {
             parser::ControlOp::Pong => Ok(()),
             parser::ControlOp::Err(e) => Err(Error::new(ErrorKind::ConnectionRefused, e)),
-            _ => Err(Error::new(ErrorKind::ConnectionRefused, "Protocol Error")),
+            other @ parser::ControlOp::Ping
+            | other @ parser::ControlOp::Msg(_)
+            | other @ parser::ControlOp::Info(_)
+            | other @ parser::ControlOp::Unknown(_) => {
```
spacejam: Even though it is unlikely these will change much for now, it's a good habit to be explicit about variants for matched enums, as it makes it easier to refactor code when you add variants. You want your code to break in relevant places when you change the variants, so that the compiler error messages point you to places where the refactor has not been accounted for yet

src/lib.rs
```diff
@@ -1051,15 +1160,15 @@ struct ServerInfo {
...
+#[inline]
 fn default_empty() -> String {
     "".to_string()
 }
-#[inline(always)]
-fn default_no_urls() -> Vec<String> {
-    vec![]
+
+const fn default_no_urls() -> Vec<String> {
+    Vec::new()
```
spacejam: Fun note: `Vec::new()` is a `const fn` because it does not actually allocate anything until you push something into the `Vec`.

src/parser.rs
```diff
@@ -53,41 +55,41 @@ impl ReadLoopState {
...
-    // FIXME(dlc) - avoid copy if possible.
-    // FIXME(dlc) - Just read CRLF? Buffered so should be ok.
-    reader
-        .take(msg_args.mlen as u64)
-        .read_to_end(&mut msg.data)?;
+        let reader = &mut self.reader;
+        // FIXME(dlc) - avoid copy if possible.
+        // FIXME(dlc) - Just read CRLF? Buffered so should be ok.
+        reader
+            .take(u64::from(msg_args.mlen))
```
spacejam: Although the clippy lints that were added will protect a lot from it, `as` is generally good to avoid where possible. It's the source of a lot of accidental truncation etc... and it's a good practice for any sort of numerical work to use `from` and `try_from` to ensure you're never truncating.

spacejam:
With the repo now being clippy-friendly, it's possible to enable this github action that will run clippy on every pull request: https://github.com/actions-rs/clippy-check

dwaynebradley:
@spacejam (Tyler), even though this is not my repo, I want to say thank you for all of the notes on the individual changes and detail you went into here. Still being a somewhat Rust "newbie", these notes are helpful learning tools for me. I greatly appreciated it.

.rustfmt.toml
```diff
@@ -0,0 +1,2 @@
+reorder_imports = true
+max_width = 100
```
spacejam: It's nice to check this into a Rust project root because people often have their own rustfmt configuration files in their home directories, and over time a lot of pull requests will then get infected with irrelevant refactors that cut long argument lines into several short ones etc... Having this in the root will override the ones that people set in their home directories sometimes, preventing their personal preferences from impacting this project's.

derekcollison:
Tyler these are great, let me settle out a few more things and will take a closer look!

src/lib.rs
```diff
@@ -443,10 +450,15 @@ fn start_flush_cycle(wbuf: &Arc<Mutex<Outbound>>) -> usize {
     w.writer.buffer().len()
 }
 
+/// A `Message` that has been published to a NATS `Subject`.
 #[derive(Debug)]
 pub struct Message {
+    /// The NATS `Subject` that this `Message` has been published to.
     pub subject: String,
+    /// The optional unique reply inbox that may be used for
```
derekcollison: Does not have to be unique, could be anything you want.

src/lib.rs
derekcollison: Good to know, I saw that they also are doing away with it.

src/lib.rs
```diff
@@ -851,8 +851,13 @@ impl Connection<Connected> {
         self.state.writer.lock().unwrap().should_flush = true;
     }
 
-    #[inline(always)]
-    fn write_pub_msg(&self, subj: &str, reply: Option<&str>, msgb: &[u8]) -> io::Result<()> {
+    #[inline]
+    fn write_pub_msg(
```
derekcollison: Just curious as to why so aggressive on line wrap?

src/lib.rs
derekcollison: On my todo list to capture and allow access. Two version, one for async server errors returned and one per subscription that has a handler for the handler returning an error.

src/lib.rs
```diff
@@ -855,12 +855,7 @@ impl Connection<Connected> {
     }
 
     #[inline]
-    fn write_pub_msg(
-        &self,
-        subj: &str,
-        reply: Option<&str>,
-        msgb: &[u8],
-    ) -> io::Result<()> {
+    fn write_pub_msg(&self, subj: &str, reply: Option<&str>, msgb: &[u8]) -> io::Result<()> {
```
derekcollison: lol - ignore my other comment.. ;)

src/parser.rs
```diff
@@ -32,13 +32,13 @@ pub(crate) struct ReadLoopState {
...
-pub(crate) fn read_loop(mut state: &mut ReadLoopState) -> io::Result<()> {
+pub(crate) fn read_loop(state: &mut ReadLoopState) -> io::Result<()> {
     loop {
         match parse_control_op(&mut state.reader)? {
-            ControlOp::Msg(msg_args) => process_msg(&mut state, msg_args)?,
+            ControlOp::Msg(msg_args) => state.process_msg(msg_args)?,
             ControlOp::Ping => state.send_pong()?,
             ControlOp::Pong => state.process_pong(),
-            _ => println!("Got something else"),
+            other => eprintln!("Received unhandled message: {:?}", other),
```
derekcollison: This was mostly for development, need to clean up and not do any prints.

src/parser.rs
```diff
@@ -53,41 +53,41 @@ impl ReadLoopState {
 
     fn send_pong(&self) -> io::Result<()> {
         let w = &mut self.writer.lock().unwrap().writer;
-        w.write(b"PONG\r\n")?;
+        w.write_all(b"PONG\r\n")?;
```
derekcollison: I don't think its needed though is it? Its a buffered writer with a flush.

src/parser.rs
```diff
@@ -53,41 +53,41 @@ impl ReadLoopState {
...
+    fn process_msg(&mut self, msg_args: MsgArgs) -> io::Result<()> {
+        let mut msg = Message {
+            subject: msg_args.subject,
+            reply: msg_args.reply,
+            data: Vec::with_capacity(msg_args.mlen as usize),
+            writer: None,
+        };
+
+        // Setup so we can send responses.
+        if msg.reply.is_some() {
```
derekcollison: Nice to know you can do this!

src/parser.rs
derekcollison: Why the extra indentation? How does behavior differ with u64::from?

src/lib.rs
derekcollison: nice!

derekcollison: I knew that but did n't think to promote to fn level.

src/lib.rs
```diff
@@ -721,9 +721,8 @@ impl Subscription {
 impl Drop for Subscription {
     fn drop(&mut self) {
         if self.do_unsub {
-            match self.unsub() {
-                Ok(_) => {}
-                Err(_) => {}
+            if let Err(error) = self.unsub() {
+                eprintln!("error unsubscribing during Subscription Drop: {:?}", error);
```
derekcollison: Not sure this makes sense, but understand since we have other println. Production lib should not do println IMO.

src/lib.rs
derekcollison: Even for buffered writers? Also does this cause large perf impact?

src/lib.rs
derekcollison: +1

src/lib.rs
```diff
@@ -1071,21 +1072,25 @@ impl Connected {
...
         Ok(())
     }
 }
 
 impl Drop for Connected {
     fn drop(&mut self) {
-        match self.close() {
-            _ => {}
+        if let Err(error) = self.close() {
+            eprintln!("error closing Connected during Drop: {:?}", error);
```
derekcollison: Not necessary, we don't care.

src/lib.rs
```diff
@@ -433,7 +433,8 @@ impl Connection<Authenticated> {
             let mut w = wbuf.lock().unwrap();
             w.in_flush = false;
             if cur_len > 0 {
-                if let Err(_) = w.writer.flush() {
+                if let Err(error) = w.writer.flush() {
+                    eprintln!("Flusher thread failed to flush: {:?}", error);
```
derekcollison: Again maybe capture for possible last_error(), eventually never println from production client lib.

src/lib.rs
derekcollison: Nice to know!

src/lib.rs
derekcollison: What is the `other @` syntax?

src/parser.rs
```diff
@@ -145,13 +145,15 @@ fn parse_msg_args(args: &[u8]) -> io::Result<ControlOp> {
         4 => (args[0], 3, Some(args[2].to_owned())),
         _ => return Err(parse_error()),
     };
-    let sid = match usize::from_str(args[1]) {
-        Ok(sid) => sid,
-        _ => return Err(parse_error()),
+    let sid = if let Ok(sid) = usize::from_str(args[1]) {
+        sid
+    } else {
+        return Err(parse_error());
```
derekcollison: I assume idiomatic, but is an extra line and does not read as well IMO, just curious.

src/lib.rs
derekcollison: Got it, thanks!

src/lib.rs
spacejam: This was actually the reason I checked in a `.rustfmt.toml` file, because I have a super aggressive personal config with a line width of 60 since I usually code with a few columns open to have as much related context directly in front of me as possible, and narrow columns are a bit faster for me to read personally.

src/parser.rs
spacejam: In general I associate wildcard matches with bugs over time. I agree that the match statement is nicer to look at.

src/lib.rs
spacejam: Yeah, if the written bytes exceed the BufWriter's inner capacity it will skip the inner buffer and go directly to the backing `Write` implementation, which only guarantees that it will return the number of bytes written. 

https://doc.rust-lang.org/src/std/io/buffered.rs.html#615

I would be pretty surprised if the perf impact caused a throughput drop. In the end, it's necessary for correct behavior, but we can avoid `BufWriter` and do all kinds of other fun stuff if that becomes an issue. If we can get that to show up on flamegraphs at all, we're doing a lot of things extremely well :)

src/lib.rs
spacejam: What do you have in mind when you say "capture"? One thing I like to do with code that runs in one thread but causes functionality associated with methods called elsewhere to misbehave is to set an AtomicPtr to refer to an error, and have all caller threads check the error pointer to see if it's null or if it points to something, which will usually be really cheap, but will feed the error asynchronously back to the callers. Are you thinking something similar?

src/lib.rs
spacejam: Thanks! I misinterpreted the docs at https://docs.nats.io/developing-with-nats/sending/replyto but looking back at it the wording is not very ambiguous that it CAN be unique instead of MUST be unique.

src/lib.rs
spacejam: I agree that it's preferable to avoid printing. What are the highest priority feedback mechanisms to support? Table stakes for rust libraries is to use the log crate, which has 0 performance impact unless the user attaches a log receiver of some sort (env_logger is what many people use as a default). Do you see things like prometheus metrics being a priority? I'm used to avoiding logs almost entirely for this stuff and using metrics instead since it keeps monitoring cheap and easy to aggregate and analyze, but the log crate is probably a good minimum thing to support that can replace this asap.

src/lib.rs
derekcollison: Gotcha.

src/lib.rs
derekcollison: For async errors from server was thinking last_error() on Connection and for callbacks for subs with handler same name but just for the handler based subscriptions.

src/lib.rs
spacejam: It's a way to bind the entire `ControlOp` enum to a name. It's useful more often when you want to destructure a complex type in a match but also have a name to refer to the overall complex type. Revisiting this particular case, it's probably better to just bind the `parser::parse_control_op(reader)?` call to a variable outside the match statement and refer to that variable from inside the passthrough arm. I'll clean that up now.

src/lib.rs
derekcollison: Yep!

src/lib.rs
derekcollison: Client libs should not print IMO.. ;)

derekcollison: But we can leave in for now, but next up want to do `last_error()` or something like it and remove all prints.

spacejam: In production people often feed it into log aggregators rather than just dumping to stderr/stdout. The nice thing about the log crate is it doesn't specify anything about how the log lines are consumed, only produced. So whatever a client wants to do with possible error messages, they will plug into the source and do whatever they want with the info. It's like NATS but for Rust library feedback. A nice way to decouple the production of messages from their consumption.

derekcollison: Agree, but client libs should only surface errors, apps can log all they want.. ;)

src/parser.rs
spacejam: The indentation comes from the fact that I moved the function that was associated with the ReadLoopState to being a method defined on it, and placed in the impl block instead of being a function outside the block.

`try_from` differs from `from` by returning a `Result` if the conversion fails. So, if you try to do `u8::try_from(u64::max_value())` it will be an `Err` because it would truncate.

derekcollison: Why don't you need the `?` at the end then?

spacejam: If using `try_from` instead of `try` you do need to handle errors (or blow up with an `expect` if it violates a core assumption that the value falls within the acceptable range). When using `u64::from` it doesn't return a `Result` but just a `u64` so there's no error to handle.

spacejam:
Thanks for the review!

src/parser.rs
derekcollison: ok so probably a naive question but why better than `as u64`?

PR #11, opened 2020-03-18
Separate complex initialization-only typestate from Connection

spacejam:
In the lead-up to implementing reconnection logic, it's important to reduce some complexity where possible to reduce the friction of introducing changes over time. I started diving into reconnection, but found myself getting a bit confused with the various typestates relating to connectivity etc... and only later did I realize they were only really being used at initialization time.

This separates concerns a bit by splitting the complex typestate into different objects that do different things: 
* `Options` which are now a publicly documented type that can be operated on in the same way that Connection could be before when it had the complex typestate. 
* `Connection` now doesn't have any complex typestate, and all access to its state has been simplified by allow it to avoid referring to the `state` member.

Personally, I'm uncertain if `Options` is the best name, but I figured I'd start with this to minimize change for now.

I've added a new file, `CHANGELOG.md` which can be used for basically copying + pasting into release notes when a new version is put out, and it documents things like breaking changes and interesting new features. I'm modeling it off of [sled's](https://github.com/spacejam/sled/blob/d6e039b24760d9a14cdf7df998f69c64b47f1cb7/CHANGELOG.md) for now, but I don't have many strong opinions about it. I mostly treat it as a way to make release notes easier by writing relevant things as they change instead of flipping through commit history since the last release.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ Basic connections, and those with options. The compiler will force these to be c
 ```rust
 let nc = nats::connect("localhost")?;
 
-let nc2 = nats::Connection::new()
+let nc2 = nats::Options::new()
     .with_name("My Rust NATS App")
     .with_user_pass("derek", "s3cr3t!")
     .connect("127.0.0.1")?;
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ Basic connections, and those with options. The compiler will force these to be c
 ```rust
 let nc = nats::connect("localhost")?;
 
-let nc2 = nats::Options::new()
+let nc2 = nats::ConnectionOptions::new()
     .with_name("My Rust NATS App")
     .with_user_pass("derek", "s3cr3t!")
     .connect("127.0.0.1")?;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -107,22 +107,22 @@ const LANG: &str = "rust";
 
 #[doc(hidden)]
 pub mod options_typestate {
-    /// `Options` typestate indicating
+    /// `ConnectionOptions` typestate indicating
     /// that there has not yet been
     /// any auth-related configuration
     /// provided yet.
     #[derive(Debug, Copy, Clone, Default)]
     pub struct Unauthenticated;
 
-    /// `Options` typestate indicating
+    /// `ConnectionOptions` typestate indicating
     /// that auth-related configuration
     /// has been provided, and may not
     /// be provided again.
     #[derive(Debug, Copy, Clone)]
     pub struct Authenticated;
 
-    /// `Options` typestate indicating that
-    /// this `Options` has been used to create
+    /// `ConnectionOptions` typestate indicating that
+    /// this `ConnectionOptions` has been used to create
     /// a `Connection` and may not be changed.
     #[derive(Debug, Copy, Clone)]
     pub struct Finalized;
@@ -130,25 +130,25 @@ pub mod options_typestate {
 
 /// A configuration object for a NATS connection.
 #[derive(Clone, Debug, Default)]
-pub struct Options<Typestate = options_typestate::Unauthenticated> {
+pub struct ConnectionOptions<TypeState = options_typestate::Unauthenticated> {
     auth: AuthStyle,
     name: Option<String>,
     no_echo: bool,
-    typestate: PhantomData<Typestate>,
+    typestate: PhantomData<TypeState>,
 }
 
-impl Options<options_typestate::Unauthenticated> {
-    /// `Options` for establishing a new NATS `Connection`.
+impl ConnectionOptions<options_typestate::Unauthenticated> {
+    /// `ConnectionOptions` for establishing a new NATS `Connection`.
     ///
     /// # Example
     /// ```
     /// # fn main() -> std::io::Result<()> {
-    /// let nc = nats::Options::new().connect("demo.nats.io")?;
+    /// let nc = nats::ConnectionOptions::new().connect("demo.nats.io")?;
     /// # Ok(())
     /// # }
     /// ```
-    pub fn new() -> Options<options_typestate::Unauthenticated> {
-        Options::default()
+    pub fn new() -> ConnectionOptions<options_typestate::Unauthenticated> {
+        ConnectionOptions::default()
     }
 
     /// Authenticate with NATS using a token.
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -23,7 +23,7 @@ Basic connections, and those with options. The compiler will force these to be c
 ```rust
 let nc = nats::connect("localhost")?;
 
-let nc2 = nats::Connection::new()
+let nc2 = nats::Options::new()
```
derekcollison: We originally had options in the Go client and then called connect on those, we wen away from that so don't want to introduce that here.

I chose the type state pattern to allow different states of a connection for compile time. For reconnect logic this will be runtime state (I have an enum) that will apply to a connected connection.

src/lib.rs
```diff
@@ -102,50 +104,20 @@ mod parser;
...
-pub trait OptionState {}
-impl OptionState for NotConnected {}
-impl OptionState for Authenticated {}
-
-// General Options
-impl<S> Connection<S>
-where
-    S: OptionState + ConnectionState,
-{
+impl Options {
```
derekcollison: You are doing a blanket one, where I also had the ability to distinguish an authorized state too which is a type of option but allows the compiler to not allow to use multiple ones.

src/lib.rs
```diff
@@ -102,50 +104,20 @@ mod parser;
...
-    pub trait Sealed {}
-    impl Sealed for super::NotConnected {}
-    impl Sealed for super::Authenticated {}
-    impl Sealed for super::Connected {}
-    impl Sealed for dyn super::OptionState {}
-}
-
-impl ConnectionState for NotConnected {}
-impl ConnectionState for Authenticated {}
-impl ConnectionState for Connected {}
```
spacejam: Sealed traits are considered to be a bit of an antipattern, unlike in the scala community where it seems they are being imported from. Leaning heavily into typestate tends to confuse contributors more than the mind-bending confusion and derailment that it can cause.

spacejam:
What is the benefit that you'd like to accomplish? There's probably a way to accomplish it without the cognitive complexity tax.

src/lib.rs
spacejam: is the change you're requesting that authentication is only settable once, in a way that is enforced at compile time?

src/lib.rs
derekcollison: Understood, but trying to optimize the end user experience and lean in on compiler support similar to serde. We could return different types like I did with SubscriptionHandler, but I liked the way docs were presented that were nice and clean and one struct, Connection.

derekcollison:
Trying to optimize end user experience and have compiler support for proper setup of a connection.

src/lib.rs
```diff
@@ -103,237 +106,129 @@ const VERSION: &str = "0.0.1";
...
-
-/// A NATS connection.
-#[derive(Debug)]
-pub struct Connection<S: ConnectionState> {
-    options: Options,
-    state: S,
+/// A configuration object for a NATS connection.
+#[derive(Clone, Debug, Default)]
+pub struct ConnectionOptions<TypeState = options_typestate::Unauthenticated> {
```
spacejam: This might look kind of weird to people, it's the specification of a default generic type. In practice it's not as helpful as it seems like it should be, but from time to time it can help reduce friction. 

One that you may have used already is std's HashMap for its 3rd generic parameter (the hashing implementation) that you can omit due to the fact that it's default.

spacejam:
I believe this addresses the requested change, what do you think, @derekcollison?

src/lib.rs
```diff
@@ -104,86 +105,127 @@ mod parser;
...
-impl Options {
-    /// Add a name option for the unconnected connection.
+#[doc(hidden)]
+pub mod options_typestate {
+    /// `Options` typestate indicating
+    /// that there has not yet been
+    /// any auth-related configuration
+    /// provided yet.
+    #[derive(Debug, Copy, Clone, Default)]
+    pub struct Unauthenticated;
```
derekcollison: Don't have a better recommendation but not a huge fan of this. Plain?

derekcollison: NoAuth?

src/lib.rs
```diff
@@ -104,86 +105,127 @@ mod parser;
...
+    /// `Options` typestate indicating that
+    /// this `Options` has been used to create
+    /// a `Connection` and may not be changed.
+    #[derive(Debug, Copy, Clone)]
+    pub struct Finalized;
+}
+
+/// A configuration object for a NATS connection.
+#[derive(Clone, Debug, Default)]
+pub struct Options<Typestate = options_typestate::Unauthenticated> {
```
derekcollison: That does look weird ;)

src/lib.rs
```diff
@@ -104,86 +105,127 @@ mod parser;
...
+    auth: AuthStyle,
+    name: Option<String>,
+    no_echo: bool,
```
derekcollison: FYI we have lots of these, the ones here were just the quick ones I chose to test out first.

src/lib.rs
```diff
@@ -104,86 +105,127 @@ mod parser;
...
+impl Options<options_typestate::Unauthenticated> {
+    /// Create a new NATS connection. This will not be a connected connection.
     ///
     /// # Example
     /// ```
     /// # fn main() -> std::io::Result<()> {
-    /// let nc = nats::Options::new()
-    ///     .with_name("My App")
-    ///     .connect("demo.nats.io")?;
+    /// let nc = nats::Options::new().connect("demo.nats.io")?;
```
derekcollison: For me this is not good, if only URL just `nats::connect(url)`

src/lib.rs
```diff
@@ -104,86 +105,127 @@ mod parser;
...
+        Options::default()
     }
 
-    /// Select option to not deliver messages that we have published.
+    /// Authenticate this NATS connection with a token.
     ///
     /// # Example
     /// ```
     /// # fn main() -> std::io::Result<()> {
     /// let nc = nats::Options::new()
```
derekcollison: nats::Options::new() seems like alot to me, you?

src/lib.rs
derekcollison: Not sure this is better than what we had to start TBH..

spacejam: Which aspect? I'm blocked on implementing reconnection logic because these refactors dramatically improve my ability to reason about the Connection object.

src/lib.rs
spacejam: This is a doc comment for Options::new, not the top-level connect function

src/lib.rs
spacejam: It's the doc for the configuration type. Doing what the std does is usually the best way to reduce friction for users: https://doc.rust-lang.org/std/fs/struct.OpenOptions.html#method.new

src/lib.rs
derekcollison: Maybe then do like fs example

`let options = nats::Options::new();`

src/lib.rs
derekcollison: That is fair.

src/lib.rs
spacejam: done

src/lib.rs
spacejam: switched to NoAuth

src/lib.rs
derekcollison: Anyway to just make Options<T>?

src/lib.rs
derekcollison: ok thanks.

src/lib.rs
derekcollison: Understood but the client seems just as complex to me now as it was before.

PR #13, opened 2020-03-23
Use concrete types to represent various iterator types

spacejam:
* using concrete types instead of impl trait allows clients to easily wrap the iterators over messages. These have been named based on the std & crossbeam's `IntoIter, Iter, TryIter` widely used types and for the timeout-based iter I renamed it `TimeoutIter` but I'm totally flexible about names
* using a lifetime to refer to a `Subscription` (or an owned version) guarantees that the `Subscription` will never drop while the iterator is still active (triggering unsub prematurely)
* implementing `IntoIterator` for both `Subscription` and `&'a Subscription` allows flexible use of `for msg in sub { .. }` syntax
* moved subscription-related logic to its own module, which keeps the new iterators from cluttering the docs, similar to how the [std does it](https://doc.rust-lang.org/std/collections/btree_map/index.html) with the `BTreeMap`'s iter method and implementations of `IntoIterator` that return one of `std::collections::btree_map::{Iter, IntoIter, IterMut}` depending on the kind of iteration underway.


spacejam:
Thanks!

PR #15, opened 2020-03-25
Refactor shared state behind a single Arc to simplify the dependency graph

spacejam:
The `SharedState` object is where I'm putting the various reconnection-related parameters. With ad-hoc `Arc`s used in different places in more fine-grained ways it's easy to cause cycles and violate assumptions about lifetimes. Now there's just one type behind a single `Arc`, simplifying the dependency graph and reducing arc-related overhead (no need to clone and store multiple `Arc`s).

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -84,6 +84,13 @@
     clippy::wrong_pub_self_convention,
 )]
 
+mod parser;
+mod shared_state;
+
+/// Functionality relating to subscribing to a
+/// subject.
+pub mod subscription;
+
 use std::{
     collections::{HashMap, VecDeque},
     io::{self, BufReader, BufWriter, Error, ErrorKind, Write},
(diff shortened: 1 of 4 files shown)


src/lib.rs
```diff
@@ -280,15 +282,18 @@ impl<TypeState> ConnectionOptions<TypeState> {
...
-            pongs: Arc::new(Mutex::new(VecDeque::new())),
-            writer: Arc::new(Mutex::new(Outbound {
-                writer: BufWriter::with_capacity(64 * 1024, stream.try_clone()?),
-                flusher: None,
-                should_flush: true,
-                in_flush: false,
-                closed: false,
-            })),
+            shared_state: Arc::new(SharedState {
+                last_error: RwLock::new(Ok(())),
```
spacejam: This can be used for surfacing errors to users, and feeding the error information through the unified state object to anything else that may be doing IO.

derekcollison: I like this, need one on subs with handlers too.

spacejam:
Thanks!

spacejam:
At some point it will likely make sense to tease this single state object apart again, but for now this is the early days of the project, and having the shared things in one place will allow things to iterate faster. But a little more discipline for separation of concerns needs to be used to prevent it from devolving into a "big ball of mud" where multiple components mutate the same shared state in ways that may cause emergent behaviors.

derekcollison:
Agree but appreciate the process so all good.

PR #16, opened 2020-03-26
reconnection logic

spacejam:
- [x] large refactor of various connection-related functionality, moving logic into the `ServerManager` code that runs on a single thread (renamed from `ReadLoopState`)
- [x] reconnection logic that tries all configured servers
- [x] allow initial configuration of several servers
- [x] truncated exponential backoff on reconnect (goes from 0ms to 4s, doubling)
- [x] resend subscriptions on reconnect
- [x] support dis/reconnection callbacks
- [x] add simple test server that occasionally drops all reconnections and facilitates pub/sub
- [x] test resubscription behavior
- [x] learn about new servers from server INFO block
- [x] test learned server reconnection based on server's INFO block
- [x] randomize servers selection on both initial connection and reconnection
- [x] make max retries configurable, falling back to using everything if everything is past the threshold
- [x] buffer outgoing messages when the connection is down
- [x] resolve hostnames to sets of IP addresses and shuffle them (unless no_shuffle is set)


kozlovic:
@spacejam Happy to discuss when you have the time. In a nutshell, the Go client when creating a connection does a hostname resolution. If multiple IPs are returned, we shuffle (unless the NoRandomize option is set) and we try to connect based on that array. We stop at the first successful tcp connection.
For the discovered servers, the initial INFO may contain a field `connect_urls` that contains a list of host:port where clients can connect to. These are added to the list of servers that the library can try to (re)connect to, although they are marked as "discovered". If the client sends `"protocol":1` in the CONNECT protocol, then servers can send an INFO asynchronously (that is, outside of the connect process) with an updated list. When processing this list, the client should remove all discovered servers that are not in this list.

spacejam:
Thanks @kozlovic :) I've added a few items to the TODO list above related to handling of specific IP addresses for a target. 

Currently implemented from what you said:
* handling async INFO and having it replace the servers from the previous INFO block 

Not yet implemented from what you said:
* tracking discovered servers
* randomizing IP addresses returned from DNS resolution

I'm currently doing these things that I'm curious about your opinion on:
1. I'm shuffling all hostnames upon initial connection and every reconnection attempt
1. If all servers are beyond the configured max_retries threshold, it will fall back to retrying everything it knows about forever

Thanks for the perspectives!

kozlovic:
> I'm shuffling all hostnames upon initial connection and every reconnection attempt

In other clients the list of servers is shuffled (if option allows) only on startup. When a server (an url really) is used, it is moved to the end of the list. This process continues on reconnecting, meaning we take the next and then move to end of the list.

When processing discovered servers, again, we remove form the array of urls the ones that are implicit (discovered), then we add the discovered ones at the end. The discovered list itself may be shuffled prior to adding them to the list of servers.

If you shuffle the whole list before each (re)connect, I think you may end-up trying the same server up to its max reconnect times while other are not even tried. Think of the possibility where the max reconnect is set to infinite (most client libs allow -1 to mean infinite).

> If all servers are beyond the configured max_retries threshold, it will fall back to retrying everything it knows about forever

Yes but see above. If you shuffle, there is always a risk that you try the same server many times before trying another that may be available. (suppose that the one you keep trying is not running at the moment). Also, I assume that you remove the ones that have reached the max from that list and when the list is empty go back to original list and now try them in sequence? Is that what you mean?

spacejam:
@kozlovic Thanks for the info! I slightly misspoke when describing the current shuffling technique. The pseudocode for the current technique is:

```
loop forever {
    randomize the url list
    filter the url list by reconnection attempts (or use all servers if that's empty)
    for each url in the shuffled and filtered url list {
        if reconnection attempts > 0 {
            sleep for truncated exponential backoff between 2ms and 4 seconds
        }
        try to connect to a server at the url
        break loops or increment reconnection count
    }
}
```

So it will try all servers at each step, in-between randomizations, preventing starvation.

I'm just storing the two sets - configured and discovered - separately to avoid any bugs relating to accidentally replacing configured ones etc...

In the past I've had huge headaches when I've failed to re-resolve dns addresses based on the provided TTL on the A record. I'm going to alter the code so that it re-resolves known hosts based on the known connect URL as well. I assume this is fine to do because it will hit the local dns cache, but please let me know if you see any issues with known deployments.

kozlovic:
@spacejam Ok.. just so you know, we have a report of a panic in the Go client because we iterate through the server pool and after processing the info and connecting ok, we try to update the "server" record based on the index of the for loop, but it looks like the pool has reduced in size during the processing of the INFO. So be aware of that.

spacejam:
@kozlovic that makes sense. I'm avoiding any explicit use of indexes, and I'm keeping the learned server sets completely separate from the configured ones to hopefully reduce bugs relating to dealing with mutating the server sets over time.

Thanks for the info!

PR #19, opened 2020-04-22
TLS Support. Closes #5.

### LINKED ISSUE #5: TLS Support
Need to decide on which tech to use and whether or not we make this a feature flag.

Contenders are:

1. [native-tls](https://crates.io/crates/native-tls)
2. [openssl](https://crates.io/crates/openssl)
3. [rustls](https://crates.io/crates/rustls)

Any thoughts and comments welcome.

spacejam:
- [x] TLS may be enabled for all outgoing connections by calling `nats::ConnectionOptions::tls_required(true)`
- [x] TLS may be enabled for a specific connection by specifying a server with the prefix `tls://host:port` rather than `nats://host:port`
- [x] TLS may be configured for all connections by using the `nats::tls::tls_connector()` method which returns a re-exported [`native_tls::TlsConnector`](https://docs.rs/native-tls/0.2.4/native_tls/struct.TlsConnectorBuilder.html) that allows various TLS settings to be configured. This applies the desired TLS configuration to all outgoing connections. Is this desired behavior?
- [x] client identity is configurable using the above `tls_connector` object

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,6 +31,7 @@ nom = "5.1"
 crossbeam-channel = "0.4"
 parking_lot = "0.10.0"
 rand = "0.7.3"
+native-tls = "0.2.4"
 
 [dev-dependencies]
 criterion = "0.3"
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -510,7 +510,7 @@ impl<TypeState> ConnectionOptions<TypeState> {
     /// # Examples
     /// ```no_run
     /// # fn main() -> std::io::Result<()> {
-    /// let mut tls_connector = nats::tls::tls_builder()
+    /// let mut tls_connector = nats::tls::builder()
     ///     .identity(nats::tls::Identity::from_pkcs12(b"der_bytes", "my_password").unwrap())
     ///     .add_root_certificate(nats::tls::Certificate::from_pem(b"my_pem_bytes").unwrap())
     ///     .build()
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -88,6 +88,7 @@ mod inbound;
 mod outbound;
 mod parser;
 mod shared_state;
+mod tls;
 
 /// Functionality relating to subscribing to a
 /// subject.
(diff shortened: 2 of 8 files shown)


src/lib.rs
```diff
@@ -234,10 +241,35 @@ impl Default for ConnectionOptions<options_typestate::NoAuth> {
...
+                if self.tls_connector.is_some() {
+                    &"set"
+                } else {
+                    &"unset"
+                },
+            )
+            .entry(&"tls_required", &self.tls_required)
+            .finish()
+    }
+}
```
spacejam, line 271: This had to be manually implemented because `native_tls::TlsConnector` unfortunately does not implement it.

src/parser.rs
```diff
@@ -114,14 +114,21 @@ fn parse_err(args: &[u8]) -> ControlOp {
...
+        reader.read_exact(byte)?;
+        buf.push(byte[0]);
+    }
+
+    if buf.starts_with(b"INFO ") {
+        if let Ok(info) = serde_json::from_slice(&buf[b"INFO ".len()..]) {
+            return Ok(info);
+        }
     }
+    Err(Error::new(ErrorKind::Other, "INFO proto not found"))
```
spacejam, line 131: This just loops over the initial INFO block one byte at a time. There are faster ways of doing this, but I assume this happens so infrequently that performance isn't really an issue here. Subsequent INFO blocks that get passed over the possibly-TLS-upgraded connection are parsed in a more efficient way.

Cargo.toml
```diff
@@ -31,6 +31,7 @@ nom = "5.1"
 crossbeam-channel = "0.4"
 parking_lot = "0.10.0"
 rand = "0.7.3"
+native-tls = "0.2.4"
```
spacejam, line 34: I went with native_tls because it allows us not to be responsible for shipping an up-to-date TLS library to our clients, as this relies on the underlying system library.

src/tls.rs
```diff
@@ -0,0 +1,92 @@
...
+        tls.flush()
+    }
+}
+
+impl TlsWriter {
+    pub(crate) fn shutdown(&mut self) -> io::Result<()> {
+        let mut tls = self.tls.lock().unwrap();
+        tls.shutdown()
+    }
+}
```
spacejam, line 92: This module includes a little trick that allows a TLS connection to be "split" into a reader and writer half. The shared TLS state machine must be held when IO is performed on the underlying socket, which means we will cause writers to block until they can acquire the mutex after a read. So, the trick here is that readers block on a call to `TcpStream::peek` on the underlying raw socket, which does not remove the bytes from the kernel's sk_buff's etc... which allows the reader to avoid holding the TLS mutex until bytes are detected on the wire. This is sort of similar to using epoll or something else to block on readiness. But this allows outgoing traffic to acqure the TLS mutex when required without preventing the reader from acquiring it when there are bytes available to be consumed.

src/parser.rs
derekcollison: I think that is fine.

src/tls.rs
derekcollison: I think that sounds good and pretty slick. Once code has settled we will want to do some perf benchmarks. I did some early on and it was pretty decent.

src/lib.rs
derekcollison: Upstream PR maybe?

src/parser.rs
```diff
@@ -114,14 +114,21 @@ fn parse_err(args: &[u8]) -> ControlOp {
...
-pub(crate) fn expect_info(reader: &mut BufReader<TcpStream>) -> io::Result<ServerInfo> {
-    let op = parse_control_op(reader)?;
+pub(crate) fn expect_info(reader: &mut TcpStream) -> io::Result<ServerInfo> {
+    let mut buf = Vec::with_capacity(512);
 
-    if let ControlOp::Info(info) = op {
-        Ok(info)
-    } else {
-        Err(Error::new(ErrorKind::Other, "INFO proto not found"))
+    while !buf.ends_with(b"\r\n") {
```
derekcollison, line 120: I see now what you did here. This might cause a hiccup in initial connect speed which we should track. Maybe add a TODO(spacejam) comment to revisit and see if we can come up with something better. We have this in alot of other clients of course and I don't think we read byte by byte. Might be worth looking quickly at Go client.

src/parser.rs
```diff
@@ -114,14 +114,21 @@ fn parse_err(args: &[u8]) -> ControlOp {
     ControlOp::Err(err_description.to_string())
 }
 
-pub(crate) fn expect_info(reader: &mut BufReader<TcpStream>) -> io::Result<ServerInfo> {
-    let op = parse_control_op(reader)?;
+pub(crate) fn expect_info(reader: &mut TcpStream) -> io::Result<ServerInfo> {
+    let mut buf = Vec::with_capacity(512);
```
derekcollison, line 118: This just initial capacity yes? Grows on demand if larger?

Cargo.toml
derekcollison: I think that sounds reasonable.

src/parser.rs
spacejam: Right, it will copy things to a larger buffer if it runs out. I saw that on my laptop with a very simple configuration it was like 283 bytes. Do you have a sense for how large this can be in larger production environments?

derekcollison: Mostly influenced by how many servers in the cluster. So if you set up a cluster with say 50 nodes (We do not recommend this BTW) you will see a larger INFO.

src/lib.rs
spacejam: I see there's been an outstanding issue since 2018, I'll give it a peek and see if it's fairly trivial https://github.com/sfackler/rust-native-tls/issues/99

derekcollison: If its easy

spacejam:
Thanks for the review!

PR #20, opened 2020-04-22
Cut version 0.3.1

spacejam:
cc @derekcollison this will let folks play with TLS right away

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+# 0.3.1
+
+## New Features
+
+* #19 TLS support has been added.
+
 # 0.3
 
 ## New Features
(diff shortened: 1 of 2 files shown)


spacejam:
#21 & #22 addressed a discrepancy with other clients where the client was spending time backing-off and retrying connections to servers that were not available.

zyc801208:
outbound.rs:115    could not compile natsio.  
when to fix this problem  version 0.3.1 ?

derekcollison:
It compiles for me, what does the out say in more detail?

spacejam:
Hey @zyc801208, I'm about to release 0.3.2 (#23) which lowers the minimum supported Rust version (MSRV) to 1.37.0. Please ensure that your Rust compiler is at least this recent. Apologies for the silent breakage! Going forward, I've added a travis test that will run on 1.37 to make sure we don't accidentally violate our MSRV as we work using the current stable compiler (the compiler keeps getting faster and faster, so it makes sense to develop against a recent version even if we target a lower version).

PR #23, opened 2020-04-23
Lower minimum supported Rust version to 1.37.0 and release 0.3.2 

spacejam:
Version 0.3.1 caused breakage for users of Rust with a compiler version 1.41 or earlier, due to the usage of the `matches!` macro which stabilized in Rust `1.41.0`. This patch does two things:
* lowers MSRV to 1.37.0
* makes our MSRV explicit

I am going to publish 0.3.2 and "yank" 0.3.1 to prevent breakage for users.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 language: rust
 rust:
+  - 1.37.0
   - stable
   - beta
   - nightly
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -487,12 +487,12 @@ impl<TypeState> ConnectionOptions<TypeState> {
     /// to apply the desired configuration to all server connections.
     ///
     /// # Examples
-    /// ```
+    /// ```no_run
     /// # fn main() -> std::io::Result<()> {
     ///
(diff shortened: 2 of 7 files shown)


src/lib.rs
```diff
@@ -487,12 +487,12 @@ impl<TypeState> ConnectionOptions<TypeState> {
     /// to apply the desired configuration to all server connections.
     ///
     /// # Examples
-    /// ```
+    /// ```no_run
     /// # fn main() -> std::io::Result<()> {
     ///
     /// let nc = nats::ConnectionOptions::new()
     ///     .tls_required(true)
-    ///     .connect("demo.nats.io")?;
```
derekcollison, line 495: The demo server has a TLS version, its just on a different port.

https://demo.nats.io:8443/varz

.travis.yml
```diff
@@ -1,5 +1,6 @@
 language: rust
 rust:
+  - 1.37.0
```
derekcollison, line 3: Do we care about going back that far?

PR #27, opened 2020-04-27
Use try instead of unwrap in doctests

spacejam:


src/lib.rs
```diff
@@ -509,12 +509,11 @@ impl<TypeState> ConnectionOptions<TypeState> {
     ///
     /// # Examples
     /// ```no_run
-    /// # fn main() -> std::io::Result<()> {
+    /// # fn main() -> Result<(), Box<dyn std::error::Error>> {
```
spacejam, line 512: this allows most errors to be converted into the return type using the try `?` operator

src/lib.rs
```diff
@@ -487,7 +487,7 @@ impl<TypeState> ConnectionOptions<TypeState> {
     /// to apply the desired configuration to all server connections.
     ///
     /// # Examples
-    /// ```
+    /// ```no_run
```
derekcollison, line 490: Why no run?

spacejam: sometimes the test fails from my home connection to the demo server. I'll run an mtr next time and find out why.

derekcollison: We must have a timeout of some sort too low. Since we are going Germany to west coast US. Let's fix that and keep running it imo.

spacejam: Loss sometimes generally just gets kind of high depending on the route. Today's weather event seems to be around the Telia backbone. Often there are weather events related to poor network maintainership of German ISPs. So, in this case it seems to be more of a normal internet weather thing. It errors out fairly frequently and tends to be a source of false-negative stress.
image

derekcollison: How long does the client try to connect currently?

PR #28, opened 2020-04-27
Add support for client_ip

spacejam:


src/inbound.rs
```diff
@@ -137,7 +137,8 @@ impl Inbound {
                 if let Ok((reader, writer, info)) = server.try_connect(&self.shared_state.options) {
                     // replace our reader and writer to correspond with the new socket
                     self.reader = reader;
-                    self.info = info;
+                    self.info = info.clone();
+                    *self.shared_state.info.write() = info;
```
derekcollison, line 141: Why do we have it in two places now?

src/shared_state.rs
```diff
@@ -239,6 +239,7 @@ pub(crate) struct SharedState {
     pub(crate) subs: RwLock<HashMap<usize, SubscriptionState>>,
     pub(crate) pongs: Mutex<VecDeque<Sender<bool>>>,
     pub(crate) outbound: Outbound,
+    pub(crate) info: RwLock<ServerInfo>,
```
derekcollison, line 242: Do we have the ability to get CID from the info as well? Or is IP first server info that client can get?

src/inbound.rs
spacejam: I was thinking it would be OK to avoid the mutex in Inbound, but that doesn't matter because this is likely to be called so infrequently from Inbound that it doesn't matter. I'll take it out of there to keep it only present in one place.

src/shared_state.rs
spacejam: IP is for whatever server the client last connected to at the time that client_ip is able to lock this mutex, and the ServerInfo is updated any time a later INFO message is received.

spacejam:
Thanks!

src/shared_state.rs
derekcollison: I meant is this the first access to the server info from the client? We probably want to surface CID as well as IP.

spacejam: https://github.com/nats-io/nats.rs/pull/30

PR #29, opened 2020-04-27
Add support for RTT calculation, as well as flush_timeout

spacejam:
I've also implemented flush_timeout to have parity with the Go client.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -821,6 +821,9 @@ impl Connection {
     }
 
     /// Flush a NATS connection by sending a `PING` protocol and waiting for the responding `PONG`.
+    /// Will fail with `TimedOut` if the server does not respond with in 60 seconds.
+    /// Will fail with `NotConnected` if the server is not currently connected.
+    /// Will fail with `BrokenPipe` if the connection to the server is lost.
     ///
     /// # Example
     /// ```
@@ -831,19 +834,13 @@ impl Connection {
     /// # }
     /// ```
     pub fn flush(&self) -> io::Result<()> {
-        // TODO(dlc) - bounded or oneshot?
-        let (s, r) = crossbeam_channel::bounded(1);
-        {
-            let mut pongs = self.shared_state.pongs.lock();
-            pongs.push_back(s);
-        }
-        self.shared_state.outbound.send_ping()?;
-        r.recv().unwrap();
-        Ok(())
+        self.flush_timeout(Duration::from_secs(60))
     }
 
     /// Flush a NATS connection by sending a `PING` protocol and waiting for the responding `PONG`.
     /// Will fail with `TimedOut` if the server takes longer than this duration to respond.
+    /// Will fail with `NotConnected` if the server is not currently connected.
+    /// Will fail with `BrokenPipe` if the connection to the server is lost.
     ///
     /// # Example
     /// ```
@@ -860,10 +857,21 @@ impl Connection {
             let mut pongs = self.shared_state.pongs.lock();
             pongs.push_back(s);
         }
+
         self.shared_state.outbound.send_ping()?;
-        r.recv_timeout(duration)
-            .map(|_| ())
-            .map_err(|_| Error::new(ErrorKind::TimedOut, "No response"))
+
+        let success = r
+            .recv_timeout(duration)
+            .map_err(|_| Error::new(ErrorKind::TimedOut, "No response"))?;
+
+        if !success {
+            return Err(Error::new(
+                ErrorKind::BrokenPipe,
+                "The connection to the remote server was lost",
+            ));
+        }
+
+        Ok(())
     }
 
     /// Close a NATS connection.
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -104,7 +104,7 @@ use std::{
         atomic::{AtomicUsize, Ordering},
         Arc,
     },
-    time::Duration,
+    time::{Duration, Instant},
 };
 
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -842,6 +842,30 @@ impl Connection {
...
+    ///
+    /// # Example
+    /// ```
+    /// # fn main() -> std::io::Result<()> {
+    /// # let nc = nats::connect("demo.nats.io")?;
+    /// nc.flush()?;
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub fn flush_timeout(&self, duration: Duration) -> io::Result<()> {
```
derekcollison, line 856: We should have a pain flush as well with default timeout like the go client.

Also, if the client gets disconnected it should release all pending flush calls with a disconnected error etc. Check Go client.

src/lib.rs
```diff
@@ -821,6 +821,9 @@ impl Connection {
     }
 
     /// Flush a NATS connection by sending a `PING` protocol and waiting for the responding `PONG`.
+    /// Will fail with `TimedOut` if the server does not respond with in 60 seconds.
```
derekcollison, line 824: Nothing should take 60 secs, even around the world for ping I would wait max 2-5 secs.. ;)

src/lib.rs
```diff
@@ -821,6 +821,9 @@ impl Connection {
...
+    /// Will fail with `NotConnected` if the server is not currently connected.
+    /// Will fail with `BrokenPipe` if the connection to the server is lost.
```
derekcollison, line 826: Maybe make this disconnected. We may not be using sockets in the future etc.

spacejam:
@derekcollison Just added a default timeout of 60 seconds to the `flush` command. Now both `flush` and `flush_timeout` will fail with `BrokenPipe` if the connection to the server is lost after beginning this request. They will fail with `NotConnected` if the server is not connected.

src/lib.rs
```diff
@@ -831,19 +834,13 @@ impl Connection {
...
-        // TODO(dlc) - bounded or oneshot?
-        let (s, r) = crossbeam_channel::bounded(1);
-        {
-            let mut pongs = self.shared_state.pongs.lock();
-            pongs.push_back(s);
-        }
-        self.shared_state.outbound.send_ping()?;
-        r.recv().unwrap();
-        Ok(())
+        self.flush_timeout(Duration::from_secs(60))
```
derekcollison, line 837: Make this like 10, I will update Go client as well.

src/lib.rs
```diff
@@ -831,19 +834,13 @@ impl Connection {
...
     }
 
     /// Flush a NATS connection by sending a `PING` protocol and waiting for the responding `PONG`.
     /// Will fail with `TimedOut` if the server takes longer than this duration to respond.
+    /// Will fail with `NotConnected` if the server is not currently connected.
+    /// Will fail with `BrokenPipe` if the connection to the server is lost.
```
derekcollison, line 843: Same as above.

src/lib.rs
derekcollison: Meaning brokenpipe may not be accurate later on.

src/lib.rs
spacejam: Ok, I copied the go client. I'll change it to 3 seconds.

derekcollison: lol yea, understand.

Do we unblock them in Go client on disconnect immediately or wait through reconnect attempts. If immediately let's do 5 secs and I will make the Go client that as well.

Maybe send an email to have all clients consistent.

spacejam:
Ok, I've updated the timeout and changed the error to ConnectionReset (the closest in name to Disconnected without going down the rabbit hole + user friction of creating our own error types).

src/lib.rs
```diff
@@ -823,7 +823,7 @@ impl Connection {
     /// Flush a NATS connection by sending a `PING` protocol and waiting for the responding `PONG`.
     /// Will fail with `TimedOut` if the server does not respond with in 10 seconds.
     /// Will fail with `NotConnected` if the server is not currently connected.
-    /// Will fail with `BrokenPipe` if the connection to the server is lost.
+    /// Will fail with `ConnectionReset` if the connection to the server is lost.
```
derekcollison, line 828: nit but we do not know yet if it has been reset, we just know we got disconnected at this point. Other client use disconnect(ed) as an event and state type.

spacejam: we're using io::Error instead of creating our own. my opinion is that it really reduces user friction (especially for network-related issues) to just use the std error type around this. https://github.com/nats-io/nats.rs/pull/29#issuecomment-620035458

spacejam: Disconnected is not an option among those: https://doc.rust-lang.org/std/io/enum.ErrorKind.html

I don't have a preference for which one gets used, but I do think it should be one of these rather than forcing users to handle our own error type.

derekcollison: I agree with using standard errors, so fair point. That is why I coded it that way and did not create our own. Thx.

derekcollison: BrokenPipe is sent back or ConnectReset? Which means by peer IIRC.

spacejam: I chose BrokenPipe because that's what happens when it's not necessarily the result of a rst/fin. ConnectionReset is a rst. Which would you prefer from https://doc.rust-lang.org/std/io/enum.ErrorKind.html?

derekcollison: Let's go with your original one, thanks for the patience.

spacejam:
Thanks for the review!

PR #32, opened 2020-04-29
Add GitHub issue templates to encourage high quality issues

spacejam:
Once we agree on the wording etc... I'm happy to repeat this for our other public repos

To see the effect of this, try opening an issue at https://github.com/rust-lang/rust

### DIFF
--- a/.github/ISSUE_TEMPLATE/blank_issue.md
+++ b/.github/ISSUE_TEMPLATE/blank_issue.md
@@ -1,0 +1,4 @@
+---
+name: Blank Issue
+about: Create an issue with a blank template.
+---
--- a/.github/ISSUE_TEMPLATE/bugs.md
+++ b/.github/ISSUE_TEMPLATE/bugs.md
@@ -1,0 +1,18 @@
+---
+name: Bug Report
+about: Report a bug found in the NATS Rust client
+labels: bug
+---
+
+Bug reports must include the following information:
+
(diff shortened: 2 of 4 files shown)


stjepang:
Suggestion for a bug report form:

----------------------------

Please fill in the following form:

1. What version of natsio was used (`grep 'name = "natsio"' Cargo.lock -A 1`)?
2. What version of Rust was used (`rustc --version`)?
3. What operating system was used? 
4. Code that behaved in an unexpected way:
```rust
// code goes here
```
5. What did you expect to happen? What happened instead?

spacejam:
@derekcollison I see that the nats-server uses a single template, and then asks the submitter to choose a box. One thing I like about the approach in this PR is that it presents people with a clean menu (similar to opening an issue on https://github.com/rust-lang/rust/issues) and then a more minimized template based on their choice.

I see that nats.go currently doesn't have an issue template, but I'm happy to add one based on the approach we decide to go with

derekcollison:
I am open here and trust you both, just want to be consistent.

spacejam:
@derekcollison definitely - I've updated the language for the bug report and feature request to more closely match the language in nats-server. If this looks good to you, I'm happy to add a similar style of issue templating to nats-server to reduce template noise for openers of either kind.

spacejam:
Thanks for the review! I'll open a similar PR for nats-server

PR #33, opened 2020-04-29
Refactor Message::respond()

stjepang:
It turns out that `shared_state` and `reply` fields in `Message` are always either both `Some` or both `None`.

In theory, it'd be great if those were a part of the same `Option`, something ilke:

```rust
reply: Option<(String, Arc<SharedState>)>,
```

However, field `reply` is already public with the `Option<String>` type so changing this would be a breaking change.

I restructured the `respond()` function to be a bit more indicative of this invariant and to fail fast if we break it.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -609,17 +609,18 @@ impl Message {
     /// # }
     /// ```
     pub fn respond(&self, msg: impl AsRef<[u8]>) -> io::Result<()> {
-        if let Some(shared_state) = &self.responder {
-            if let Some(reply) = &self.reply {
-                shared_state.outbound.send_response(reply, msg.as_ref())?;
+        match (&self.responder, &self.reply) {
+            (Some(shared_state), Some(reply)) => {
+                shared_state.outbound.send_response(reply, msg.as_ref())
             }
-        } else {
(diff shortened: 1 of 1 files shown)


stjepang:
@spacejam CI fails with:

```
running 1 test
test simple_reconnect ... test simple_reconnect has been running for over 60 seconds
```

Is this a known issue?

derekcollison:
Not that I am aware, but right now uses the demo server IIRC which is in Oregon and Travis is in Germany. I will be expanding the demo network at some point so I like this approach.

spacejam:
@stjepang I believe it's failing due to a real race condition right now, which sometimes gets hit on travis but I haven't been able to repro yet on my laptop. I think it's not worth spending time on until we discuss the planned architecture changes (single threaded + smol) so we avoid debugging work that fixes bugs related to a deadlock somewhere that will simply go away when we shift to a better architecture.

spacejam:
I'm going to merge this because the issue is with code that is not relevant to this particular PR. We may want to open another issue to track it - but it will be going away soon due to architecture changes that reduce the risk of any race conditions creeping in due to the current mutex network.

PR #34, opened 2020-04-30
Address bug with PONG receiver leaks during disconnection

spacejam:
Just realized this bug exists while scanning the source code

Bug:
* client is disconnected
* client calls `Connection::flush`
  * a `crossbeam_channel::Sender` is pushed into the pongs `VecDeque`
  * `Outbound::send_ping` is called, which returns `Err(NotConnected)`
  * the client early-exits from `Connection::flush`
* client reconnects
* client calls `Connection::flush` again, adds new `Sender` to pong queue, sends PING to server
* server sends back PONG
* `Inbound` receives and parses the PONG, pops the first `Sender` from the pong queue (which is no longer waiting)
* the second client call hangs on the crossbeam_channel's Receiver until the receive timeout fires, when it should have actually received a success.

We need to hold the pong queue mutex across the call to send_ping (which may fail) because we must linearize the items in the queue with calls to send_ping. Only holding it on one end of this operation will race other calls to flush as well as transitions into the disconnected state.

We need to hold the pong queue mutex across calls to `Outbound::transition_to_disconnect_buffer` because otherwise, it may not synchronize properly with the calls to `flush`.

@stjepang and I are thinking about a simpler, lower-mutex architecture that we can shift to after implementing nkeys, where only a single thread controls a simple state machine and basically guarantees bugs like this are impossible by handling events in a serialized way.

### DIFF
--- a/src/inbound.rs
+++ b/src/inbound.rs
@@ -86,17 +86,28 @@ impl Inbound {
     }
 
     fn reconnect(&mut self) -> io::Result<()> {
+        // we must hold this mutex while changing state to disconnected,
+        // setting the outbound buffer to disconnected, and then clearing
+        // all in-flight pongs.
+        let mut pongs = self.shared_state.pongs.lock();
+
         self.status = ConnectionStatus::Disconnected;
+
+        // we must call this while holding the pongs lock to ensure that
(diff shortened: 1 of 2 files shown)


spacejam:
Thanks for the review!

PR #35, opened 2020-04-30
Add authentication with credentials file

stjepang:
This PR adds `ConnectionOptions::with_credentials()` that allows you to specify a `.cred` file to authenticate.

Connection options handling is becoming unwieldy - would you be okay with me doing a little bit refactoring around that? In particular, I'd like to move everything options-related into a dedicated file.

### DIFF
--- a/src/connect.rs
+++ b/src/connect.rs
@@ -81,7 +81,7 @@ fn is_empty_or_none(field: &Option<String>) -> bool {
     }
 }
 
-/// Attempts to connect to a server using a single socket addrs.
+/// Attempts to connect to a server on a single `SocketAddr`.
 pub(crate) fn connect_to_socket_addr(
     addr: SocketAddr,
     host: String,
@@ -93,44 +93,14 @@ pub(crate) fn connect_to_socket_addr(
     let mut stream = TcpStream::connect(&addr)?;
     let server_info = crate::parser::expect_info(&mut stream)?;
 
-    let connect_info = authenticate(server_info.clone(), options.clone(), tls_required)?;
+    let (mut reader, writer) = authenticate(
+        stream,
+        server_info.clone(),
+        options.clone(),
+        tls_required,
+        host.clone(),
+    )?;
 
-    let op = format!(
-        "CONNECT {}\r\nPING\r\n",
-        serde_json::to_string(&connect_info)?
-    );
-
-    // potentially upgrade to TLS
-    let (mut reader, mut writer) =
-        if options.tls_required || server_info.tls_required || tls_required {
-            let attempt = if let Some(ref tls_connector) = options.tls_connector {
-                tls_connector.connect(&host, stream)
-            } else {
-                match native_tls::TlsConnector::new() {
-                    Ok(connector) => connector.connect(&host, stream),
-                    Err(e) => return Err(io::Error::new(io::ErrorKind::Other, e)),
-                }
-            };
-            match attempt {
-                Ok(tls) => {
-                    let (tls_reader, tls_writer) = split_tls(tls);
-                    let reader = Reader::Tls(BufReader::with_capacity(64 * 1024, tls_reader));
-                    let writer = Writer::Tls(BufWriter::with_capacity(64 * 1024, tls_writer));
-                    (reader, writer)
-                }
-                Err(e) => {
-                    log::error!("failed to upgrade TLS: {:?}", e);
-                    return Err(io::Error::new(io::ErrorKind::PermissionDenied, e));
-                }
-            }
-        } else {
-            let reader = Reader::Tcp(BufReader::with_capacity(64 * 1024, stream.try_clone()?));
-            let writer = Writer::Tcp(BufWriter::with_capacity(64 * 1024, stream));
-            (reader, writer)
-        };
-
-    writer.write_all(op.as_bytes())?;
-    writer.flush()?;
     let parsed_op = parse_control_op(&mut reader)?;
 
     match parsed_op {
@@ -150,10 +120,12 @@ pub(crate) fn connect_to_socket_addr(
 }
 
 fn authenticate(
+    stream: TcpStream,
     server_info: ServerInfo,
     options: FinalizedOptions,
     tls_required: bool,
-) -> io::Result<ConnectInfo> {
+    host: String,
+) -> io::Result<(Reader, Writer)> {
     // This regex parses a credentials file.
     //
     // The credentials file is typically `~/.nkeys/creds/synadia/<account/<account>.creds` and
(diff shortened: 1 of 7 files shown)


src/shared_state.rs
```diff
@@ -370,3 +416,11 @@ fn empty_or_none(field: &Option<&String>) -> bool {
         None => true,
     }
 }
+
+#[inline]
+fn empty_or_none_owned(field: &Option<String>) -> bool {
```
stjepang, line 421: Would be better if we didn't need this function. The fact that `Connect<'_>` has a lifetime and is borrowing temporary variables makes the code a bit more convoluted that it needs to be, I think.

What if instead of having fields of type `Option<&String>` we simply had `Option<String>` (talking about `user`, `pass`, `auth_token`, etc)? It's not like saving a few allocations in the options struct saves us any tangible performance...

src/shared_state.rs
```diff
@@ -111,13 +123,30 @@ impl Server {
...
 
-        let op = format!(
-            "CONNECT {}\r\nPING\r\n",
-            serde_json::to_string(&connect_op)?
-        );
+                let key_pair = KeyPair::from_seed(&seed)
+                    .map_err(|err| Error::new(ErrorKind::InvalidData, err))?;
+                nkey = Some(key_pair);
+
+                jwt = Some(user_jwt);
```
derekcollison, line 145: Where do we sign the nonce?

src/lib.rs
```diff
@@ -557,9 +586,16 @@ pub struct Connection {
...
 enum AuthStyle {
-    //    Credentials(String, String),
+    /// Authenticate using a token.
     Token(String),
+
+    /// Authenticate using a username and password.
     UserPass(String, String),
+
+    /// Authenticate using a `.creds` file.
+    Credentials(PathBuf),
```
derekcollison, line 596: This in the Go client is simply and concatenation of JWT and pl(nkey) which we will want to also support as separate ones. Look at Go client for pointers.

src/shared_state.rs
derekcollison: What we want ideally is that this is loaded on demand only when needed to form CONNECT {}.

src/shared_state.rs
```diff
@@ -161,11 +190,24 @@ impl Server {
...
+        nkey: Option<&KeyPair>,
     ) -> io::Result<(Reader, Writer, ServerInfo)> {
         let mut stream = TcpStream::connect(&addr)?;
         let info = crate::parser::expect_info(&mut stream)?;
 
+        if !info.nonce.is_empty() {
+            let nkey = nkey.unwrap();
+            let sig = nkey.sign(info.nonce.as_bytes()).unwrap();
+            let sig = base64_url::encode(&sig);
+            connect_op.sig = Some(sig);
```
derekcollison, line 203: Try to wipe memory as best as we can once the CONNECT is sent.

src/shared_state.rs
derekcollison: Makes sense.

src/shared_state.rs
spacejam: Implemented buffer & string zeroing via the new `SecureVec` & `SecureString` structs that scramble their data on Drop

src/lib.rs
stjepang: Done - this variant now looks like this:
```rust
/// Authenticate using a `.creds` file.
Credentials {
    /// Securely loads the user JWT.
    jwt_cb: Arc<dyn Fn() -> io::Result<String> + Send + Sync>,
    /// Securely loads the nkey and signs the nonce passed as an argument.
    sig_cb: Arc<dyn Fn(&[u8]) -> io::Result<SecureString> + Send + Sync>,
},
```

src/lib.rs
```diff
@@ -349,6 +349,35 @@ impl ConnectionOptions<options_typestate::NoAuth> {
...
     }
+
+    /// Authenticate with NATS using a credentials file
+    ///
+    /// # Example
+    /// ```no_run
+    /// # fn main() -> std::io::Result<()> {
+    /// let nc = nats::ConnectionOptions::new()
+    ///     .with_credentials("path/to/my.creds")
+    ///     .connect("demo.nats.io")?;
```
derekcollison, line 365: Make `connect.ngs.global`

src/lib.rs
derekcollison: Looks good!

stjepang:
Okay, this is ready for review now. Changes:

- We added secure wiping of sensitive data.
- Refactored the code a bit to untangle the connection process and added more comments.
- Modeled `Credentials` after the Go client - now we have two callbacks in it - one to get the user JWT and one to sign a nonce.

derekcollison:
Looks good.

src/secure_wipe.rs
```diff
@@ -0,0 +1,104 @@
...
+        let mut rng = thread_rng();
+        for byte in &mut self.0 {
+            #[allow(unsafe_code)]
+            unsafe {
+                write_volatile(byte, rng.gen());
+            }
+        }
+        compiler_fence(SeqCst);
+    }
+}
```
spacejam, line 36: this is broken without the push function that I included originally

src/connect.rs
```diff
@@ -0,0 +1,207 @@
...
+    /// The version of the client.
+    pub version: String,
+
+    /// Indicates whether the client requires an SSL connection.
+    #[serde(default)]
+    pub tls_required: bool,
+
+    /// Connection username (if `auth_required` is set)
+    #[serde(skip_serializing_if = "is_empty_or_none")]
+    pub user: Option<String>,
```
spacejam, line 54: `SecureString`

src/connect.rs
```diff
@@ -0,0 +1,207 @@
...
+pub(crate) struct ConnectInfo {
+    /// Turns on +OK protocol acknowledgements.
+    pub verbose: bool,
+
+    /// Turns on additional strict format checking, e.g. for properly formed subjects.
+    pub pedantic: bool,
+
+    /// User's JWT.
+    #[serde(rename = "jwt", skip_serializing_if = "is_empty_or_none")]
+    pub user_jwt: Option<String>,
```
spacejam, line 26: SecureString

src/connect.rs
```diff
@@ -0,0 +1,207 @@
...
+
+    /// Signed nonce, encoded to Base64URL.
+    #[serde(rename = "sig", skip_serializing_if = "secure_is_empty_or_none")]
+    pub signature: Option<SecureString>,
```
spacejam, line 30: SecureString

src/connect.rs
```diff
@@ -0,0 +1,207 @@
...
+        auth_token: None,
+        user_jwt: None,
+        signature: None,
+        echo: !options.no_echo,
+    };
+
+    // Fill in the info that authenticates the client.
+    match &options.auth {
+        AuthStyle::UserPass(user, pass) => {
+            connect_info.user = Some(user.to_string());
```
spacejam, line 154: SecureString

src/secure_wipe.rs
spacejam: when parser::expect_info reads in the ServerInfo, it may realloc, causing non-scrambled copies of memory to float around in the vacated allocations

PR #39, opened 2020-05-05
Fix the hardcoded client version

stjepang:
Now the client version number will always be correct.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -132,7 +132,7 @@ use {
     tls::{split_tls, TlsReader, TlsWriter},
 };
 
-const VERSION: &str = "0.0.1";
+const VERSION: &str = env!("CARGO_PKG_VERSION");
 const LANG: &str = "rust";
 
 /// Information sent by the server back to this client


stjepang:
Yes - this is how everybody is using the package constant as far as I can tell.

Just to be sure, I created a simple project, added `natsio` as a dependency, and then printed the version on the screen. The reported version is one of `natsio`, not my project, which is good.

@spacejam Could I be missing any gotchas here?

derekcollison:
Can you compile a rust app against our client lib without cargo? Not saying you should, but can you? And if so what gets inserted, just empty string?

spacejam:
@derekcollison you can technically do this, but the few folks who do this are quite aware of the many, many issues they are likely to encounter by doing so, and will know to grep for things like this and account for them in forks that they maintain themselves.

derekcollison:
ok just needed to check. Seen this goes sideways a few times.

stjepang:
@derekcollison If somebody is compiling this library without cargo, compilation fails because `CARGO_PKG_VERSION` isn't set.

In other words, the version string will not get silently set to some garbage if compiling without Cargo. And if you're manually setting this environment variable, you'll likely have bigger problems to deal with :)

You can see this by playing with the following example: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=288dd9eca1bdf0725a802f35299d8607

PR #41, opened 2020-05-06
Add reconnection jitter and the ability to set a callback that calculates the reconnect delay

spacejam:


src/lib.rs
```diff
@@ -516,9 +516,15 @@ impl<TypeState> ConnectionOptions<TypeState> {
         self
     }
 
-    /// Set a callback to be executed when the client has been
-    /// closed due to exhausting reconnect retries to known servers
-    /// or by completing a drain request.
+    /// Set a callback to be executed for calculating the backoff duration
+    /// to wait before a server reconnection attempt.
```
derekcollison, line 575: cluster, since it makes it clear that this callback should fire only after all servers were not able to connect.

src/shared_state.rs
```diff
@@ -123,19 +123,21 @@ impl Server {
...
+            (cb)(self.reconnects)
+        } else if self.reconnects > 0 {
             let log_2_four_seconds_in_ms = 12_u32;
             let truncated_exponent = std::cmp::min(
                 log_2_four_seconds_in_ms,
                 u32::try_from(std::cmp::min(u32::max_value() as usize, self.reconnects)).unwrap(),
             );
-            2_u64.checked_pow(truncated_exponent).unwrap()
+
+            let jitter = thread_rng().gen_range(0, 1000);
```
derekcollison: This always fixed to 1ms? Might be reading it wrong.

src/lib.rs
```diff
@@ -241,6 +241,21 @@ pub struct ConnectionOptions<TypeState> {
...
 
+fn default_reconnect_delay_callback(reconnects: usize) -> Duration {
+    if reconnects > 0 {
+        let log_2_four_seconds_in_ms = 12_u32;
+        let truncated_exponent = std::cmp::min(
+            log_2_four_seconds_in_ms,
+            u32::try_from(std::cmp::min(u32::max_value() as usize, reconnects)).unwrap(),
+        );
+
+        let jitter = thread_rng().gen_range(0, 1000);
```
derekcollison, line 257: Same here.

src/shared_state.rs
spacejam: this is between 0 and 1000ms, which is the same as the Go client with up to 1s of default jitter

PR #46, opened 2020-05-13
Make multiple server specification whitespace resistant

spacejam:


spacejam:
thanks for the review!

PR #47, opened 2020-05-15
Allow --creds flag

derekcollison:
I tried several versions of this, feels like it should be easier.

First attempt always called with_credentials() and wanted it to do right thing with empty creds.

Second attempt built up options, but when I tried top reassign with credentials complained because it was different types.

Signed-off-by: Derek Collison <derek@nats.io>

### DIFF
--- a/examples/nats-box/main.rs
+++ b/examples/nats-box/main.rs
@@ -8,6 +8,9 @@ struct Cli {
     /// NATS server
     #[structopt(long, short, default_value = "demo.nats.io")]
     server: String,
+    // Optional credentials for NATS 2.x
+    #[structopt(long, default_value = "")]
+    creds: String,
     /// Command: pub, sub, request, reply
     #[structopt(subcommand)]
     cmd: Command,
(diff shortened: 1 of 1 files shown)


examples/nats-box/main.rs
```diff
@@ -28,12 +31,24 @@ enum Command {
 
 fn main() -> CliResult {
     let args = Cli::from_args();
-    let nc = nats::ConnectionOptions::new()
-        .with_name("nats-box rust example")
-        .connect(&args.server)?;
+    let nc: nats::Connection;
+
+    if args.creds.len() > 0 {
```
spacejam, line 36: can creds be an `Option<String>` to avoid conflating emptiness with presence?

derekcollison: Not sure, depends on CLI framework.

Regardless of that, what would your optimum connection constructor look like here with optional creds?

spacejam: I'd recommend doing this: https://github.com/nats-io/nats.rs/blob/3c8e91690d337f3cbe499e5a48a46288216cae0d/examples/nats_bench.rs#L19-L21

derekcollison: That is why I asked, I like it! I will make the changes.. Thanks!

derekcollison: Please take another look, thanks again for the suggestion, exactly what I was looking for.

PR #51, opened 2020-05-19
Name all threads created for normal system operation

spacejam:
Now that performance is becoming a subject of conversation, this will surface information to profilers to improve output

(screenshot from using [hotspot](https://github.com/KDAB/hotspot) w/ data generated from [cargo flamegraph](https://github.com/flamegraph-rs/flamegraph) with the command `cargo flamegraph --example nats_bench -- test -n 10000000` using the new bench tool.

#### before

before

#### after
after

### DIFF
--- a/src/shared_state.rs
+++ b/src/shared_state.rs
@@ -200,13 +200,13 @@ impl SharedState {
         };
 
         let inbound_thread = thread::Builder::new()
-            .name("nats_inbound_read_loop".to_string())
+            .name(format!("nats_inbound_{}", shared_state.id))
             .spawn(move || inbound.read_loop())
             .expect("threads should be spawnable");
 
         let outbound_state = shared_state.clone();
         let outbound_thread = thread::Builder::new()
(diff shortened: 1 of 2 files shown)


src/shared_state.rs
```diff
@@ -199,10 +199,16 @@ impl SharedState {
             shared_state: shared_state.clone(),
         };
 
-        let inbound_thread = thread::spawn(move || inbound.read_loop());
+        let inbound_thread = thread::Builder::new()
+            .name("nats_inbound_read_loop".to_string())
+            .spawn(move || inbound.read_loop())
+            .expect("threads should be spawnable");
```
spacejam, line 205: while this looks like it's adding a new panic, thread::spawn will panic in the same way by default. I figure that since we're about to move to the single-thread model, we don't need to break the API to avoid panicking, as this will only be in place for a short while.

spacejam:
@derekcollison yeah, I'll add those to the thread names, thanks!

PR #53, opened 2020-05-19
Auto-tune flusher back-off

spacejam:
this brings the rust client more-or-less up to the go client's performance.

on my laptop, this gets us to 6.5m writes/s (before this is 2.8m/s due to heavy lock contention and the flusher constantly writing small buffers), with the go client doing 7.4m/s.

### DIFF
--- a/src/outbound.rs
+++ b/src/outbound.rs
@@ -153,8 +153,8 @@ impl Outbound {
 
     pub(crate) fn flush_loop(&self) {
         inject_delay();
-        let mut writer = self.writer.lock();
         loop {
+            let mut writer = self.writer.lock();
             while writer.flusher_should_wait() {
                 self.updated.wait(&mut writer);
             }
@@ -163,6 +163,8 @@ impl Outbound {
                 log::info!("flusher thread shutting down");
                 return;
             }
+
+            let before = std::time::Instant::now();
 
             if let Err(error) = writer.flush() {
                 log::error!("Outbound thread failed to flush: {:?}", error);
@@ -173,6 +175,10 @@ impl Outbound {
                 // replaces our buffer
                 self.updated.wait(&mut writer);
             }
+
+            drop(writer);
+
+            std::thread::sleep(before.elapsed());
         }
     }
 


src/outbound.rs
```diff
@@ -153,8 +153,8 @@ impl Outbound {
 
     pub(crate) fn flush_loop(&self) {
         inject_delay();
-        let mut writer = self.writer.lock();
         loop {
+            let mut writer = self.writer.lock();
             while writer.flusher_should_wait() {
                 self.updated.wait(&mut writer);
```
derekcollison, line 159: Do we drop lock inside of here?

src/outbound.rs
```diff
@@ -173,6 +175,10 @@ impl Outbound {
                 // replaces our buffer
                 self.updated.wait(&mut writer);
             }
+
+            drop(writer);
+
+            std::thread::sleep(before.elapsed());
```
derekcollison, line 181: Should we sleep here?

src/outbound.rs
spacejam: no, because this is a parking_lot mutex and condvar rather than a std one that takes the owned one. the compiler guarantees that we hold the mutex while accessing the writer held inside later on in the loop

src/outbound.rs
spacejam: we want to allow writers to have a chance to batch up their writes a bit, which they won't be able to do if we keep taking out the mutex in a loop. but the issue we want to avoid is sleeping longer than necessary, which would add too much latency to the non-loaded system.

this approach aims to maintain roughly 50% lock time dedicated to the flusher, and 50% of the time available for the user to batch up writes. so, it avoids sleeping too long, and auto-tunes to roughly maintain this proportion of writer-related contention.

derekcollison: Should we sleep first thing when kicked?

src/outbound.rs
derekcollison: Ah yes, I think I had that in original just did not recognize..

src/outbound.rs
spacejam: this is roughly based on a batching technique that @utaal mentioned to me for auto-tuning batch sizes in extremely high-throughput distributed dataflow processing systems, while still maintaining nice latency for messages while under low-load

spacejam: if we sleep right after being kicked, we double the mutex-related contention because we double-lock on each acquisition

PR #57, opened 2020-06-03
Implement drain functionality for the Subscription and Connection

spacejam:


src/outbound.rs
```diff
@@ -263,6 +263,66 @@ impl Outbound {
...
+        pong_rx: crossbeam_channel::Receiver<bool>,
+    ) -> io::Result<()> {
+        // take lock, which will be held for the duration
+        // of this operation
+        let mut writer = self.writer.lock();
+
+        if writer.is_disconnected() {
+            return Err(Error::new(
+                ErrorKind::NotConnected,
+                "The client is not currently connected to a server",
```
derekcollison, line 278: Make consistent with below, so maybe "the connection is not currently connected"

src/outbound.rs
```diff
@@ -263,6 +263,66 @@ impl Outbound {
...
+    pub(crate) fn drain(
+        &self,
+        sids: Vec<usize>,
+        pong_rx: crossbeam_channel::Receiver<bool>,
+    ) -> io::Result<()> {
+        // take lock, which will be held for the duration
+        // of this operation
+        let mut writer = self.writer.lock();
+
+        if writer.is_disconnected() {
```
derekcollison, line 275: Should this error or just return?

src/outbound.rs
```diff
@@ -263,6 +263,66 @@ impl Outbound {
...
+        }
+
+        // send PING
+        writer.write_all(b"PING\r\n")?;
+
+        // Flush in place on pings.
+        writer.flush()?;
+
+        // wait for PONG
+        let success = pong_rx
```
derekcollison, line 301: This take into account if other pings are oustanding?

src/outbound.rs
```diff
@@ -263,6 +263,66 @@ impl Outbound {
...
+        for sid in sids {
+            write!(writer, "UNSUB {}\r\n", sid)?;
+        }
+
+        // send PING
+        writer.write_all(b"PING\r\n")?;
+
+        // Flush in place on pings.
+        writer.flush()?;
+
```
derekcollison, line 299: Have to double check Go client but I believe after this point we want to disallow publish() calls.

src/outbound.rs
spacejam: well, if they are calling drain for some assurance that they are receiving things, then we may not want to give them that assurance if the system is in a state of decreased availability. but I don't know what the expectations of users are. If people call drain without any hope that the server is in communication then a return is OK.

src/outbound.rs
spacejam: that is implicit in the writer lock being held

src/outbound.rs
spacejam: pongs are serialized through the pongs mutex protecting the VecDeque of senders for them, guaranteeing that the one pushed in by the caller of this function is in-order and corresponds to the above ping.

src/outbound.rs
spacejam: which one below? the one directly below is only called when the system is in a permanently closed state, rather than in a temporarily disconnected state. Should the system immediately move to closed if it's disconnected, instead of doing a flush?

src/outbound.rs
derekcollison: If they are disconnected there is nothing to drain ;)

src/outbound.rs
derekcollison: One referenced connection, the other client. Just wanted them consistent.

src/outbound.rs
spacejam: what if they just published some things into a topic that they are also subscribing to, as a way of seeing round-trips? shouldn't they get an error if they're calling drain as a kind of "barrier" operation the way that they would call "flush"?

PR #62, opened 2020-06-18
Derive Clone for Message

bfrog:
I'd like to be able to clone() this type as I've refactored out previous usage of natsclient in a project which derived clone on message. It makes it easier to clone the message to be passed around to multiple threads


spacejam:
LGTM, the error that this encountered was due to a dependency silently raising its MSRV to 1.40, and we have raised ours up to 1.40 for that reason and another in #63

PR #63, opened 2020-06-24
Bump MSRV to 1.40.0

spacejam:


spacejam:
This addresses the breakage that happened due to the base64 library using `cfg(doctest)` which was stabilized in 1.40.0. This also addresses the breakage encountered in https://github.com/nats-io/nats.rs/pull/43#issuecomment-648508852 due to some async compiler functionality still not being in place.

PR #64, opened 2020-06-24
Use GitHub Actions for all testing

spacejam:


spacejam:
travis: 
image

gha:
image

more tests in a small fraction of the time = huge developer experience improvement

PR #65, opened 2020-06-29
Cut version 0.6.0

spacejam:


spacejam:
This needs to be a new 0.6.0 version instead of 0.5.1 because we have bumped the MSRV from 1.39 to 1.40

PR #68, opened 2020-06-29
Use async sleeps in debug delay to speed up fault injection test

spacejam:


spacejam:
going to merge this since it's a testing-only change

PR #69, opened 2020-06-30
Initial headers implementation

spacejam:
the doctest passes when used against a local recent server running the master branch of nats-server, but fails against demo.nats.io. @derekcollison the doctest is no_run for now until `demo.nats.io` uses our next release that includes header support.

### DIFF
--- a/src/asynk/proto.rs
+++ b/src/asynk/proto.rs
@@ -1,10 +1,11 @@
+use std::convert::TryFrom;
 use std::io::{self, Error, ErrorKind};
 use std::str::FromStr;
 
 use futures::prelude::*;
 
 use crate::connect::ConnectInfo;
-use crate::{inject_io_failure, ServerInfo};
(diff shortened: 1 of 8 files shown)


derekcollison:
Will update demo server to more recent version.

stjepang:
@spacejam Looks like you forgot to `git add src/headers.rs` ;)

src/asynk/proto.rs
```diff
@@ -123,6 +134,94 @@ pub(crate) async fn decode(mut stream: impl AsyncBufRead + Unpin) -> io::Result<
         }));
     }
 
+    if line_uppercase.starts_with("HMSG") {
+        // Extract whitespace-delimited arguments that come after "MSG".
```
stjepang, line 138:
```suggestion
        // Extract whitespace-delimited arguments that come after "HMSG".
```

PR #70, opened 2020-06-30
Switch from native-tls to rustls

### LINKED ISSUE #61: Switch from native-tls to rustls
As someone who has lost a cumulative total of 37 hours (yes, I've been keeping track) every time I've been stalled by a rust build chain that required the `native-tls` crate, which in turn requires the abomination known as `openssl-sys`, I would love for my use of the NATS client to not give its consumers this kind of headache.

Every time you build the `native-tls` crate in any kind of cross-platform scenario, you have to do so on a machine that has the open SSL source code/header libraries installed for the right target platform and OS and trust me when I say that even having this installed doesn't mean it'll work.

**tl;dr** - cross-platform/multi-platform build pipelines become insufferably difficult in the presence of anything that relies on openssl. The `rustls` crate does not have this problem.

stjepang:
Public API changes:

- Delete `tls` module.
- Delete `Options::tls_connector()` method.
- Add `Options::add_root_certificate()` method.

While this is a breaking change, the new API is simpler and allows us to swap out TLS implementation in the future without affecting the public API.

Closes #61 
cc @autodidaddict


spacejam:
side-note, it's nice to have all tests go green in 2 minutes now :)

autodidaddict:
I'm using the latest stuff but I haven't connected over tls to servers in a while as I've been doing local host testing

derekcollison:
Works fine for me from the client repo.

`cargo run --example nats-box -- -s tls://demo.nats.io:4443 pub foo "HELLO RUST"`

PR #72, opened 2020-07-01
Revert rustls

spacejam:
This reverts commit a8a265a05e9f09775b878877d9ab387634bc6714, reversing
changes made to fb417be755a49b9ae3766ae60797d386ce168db5.

cc @stjepang


derekcollison:
I commented in Slack, need to add in the webpki crate from mozilla and add the root certificates in there. Or at least we should add in NGS etc.

PR #78, opened 2020-07-09
Update dependencies

stjepang:
The async-tls version we're using was yanked, see: https://github.com/async-rs/async-tls/issues/26

I also took this chance to update all of our dependencies.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -26,7 +26,9 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, macos-latest, windows-latest]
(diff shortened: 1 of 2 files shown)


derekcollison:
Should we do a release on this one?

stjepang:
Had to disable windows-latest platform in CI because there's an unrelated issue with rustup. We'll re-enable it when that gets fixed.

@derekcollison Yes, it's time for a new release. @spacejam suggests we do it on Monday so that we can fix issues promptly, if any occur.

dwaynebradley:
I was getting ready to submit an issue about this since I ran into it earlier today running against the master branch. 😉 Glad a fix already done!

PR #81, opened 2020-07-16
Add auth-token flag. Group auth-token and certs into auth clap::ArgGroup

xoac:
Just a small improvement to nats-box. It's really useful on embedded devices. Thank you.


spacejam:
LGTM, thank you :)

PR #83, opened 2020-07-21
Fix a deadlock when draining subscriptions

### LINKED ISSUE #82: subscription close does not end the subscription iterator in v0.6 and v0.7

 - [ x] Included below version and environment information
 - [ x] Included a [Minimal, Complete, and Verifiable example] (https://stackoverflow.com/help/mcve)

#### natsio version (`grep 'name = "natsio"' Cargo.lock -A 1`)
version 0.7

#### rustc version (`rustc --version` - we support Rust 1.37 and up)
1.45

linux (fedora 31)
```
use std::{thread, time::Duration};

fn read_loop(subscription: nats::Subscription) {
    println!("reading messages");
    for msg in subscription.messages() {
        println!("{:?}", msg);
    }
    println!("subscription closed");
}

fn main() {
    let nats_conn = nats::connect("127.0.0.1").unwrap();

    let mut subscription = nats_conn.subscribe("some-channel").unwrap();

    let s1 = subscription.clone();
    let thread_wait = thread::Builder::new().spawn(move || read_loop(s1)).unwrap();

    thread::sleep(Duration::from_millis(2000));
    println!("stopping subscription");
    subscription.drain().unwrap();

    println!("waiting for thread");
    thread_wait.join().unwrap();
    println!("done waiting");
}
```
#### Expected result:
Expected the program to shutdown within a few seconds.

#### Actual result:
Program hangs at subscription.drain() line. Can reproduce it in v0.6 and v0.7. Works in v0.5.

replacing subscription.drain() with subscription.close() also has the same issue.

stjepang:
Since `Subscription` is cloneable and `asynk::Subscription` isn't, `Subscription` would contain an `Arc<Mutex<asynk::Subscription>>` inside it.

A deadlock occurs if a `next()` operation acquires the mutex and then yields because there are no messages to receive. Then, a `drain()` operation would not be able to acquire the mutex so it would yield too without ever waking up the `next()` operation.

The solution is to make `asynk::Subscription` shareable across threads (by making its methods require `&self` rather than `&mut self`). Then we don't need the mutex at all and everything becomes much simpler. Really, this PR is more of a cleanup than a bugfix :)

Fixes #82

### DIFF
--- a/src/asynk/subscription.rs
+++ b/src/asynk/subscription.rs
@@ -2,6 +2,7 @@ use std::fmt;
 use std::io;
 use std::pin::Pin;
 use std::task::{Context, Poll};
+use std::sync::atomic::{AtomicBool, Ordering};
 
 use blocking::block_on;
 use futures::prelude::*;
@@ -24,7 +25,7 @@ pub struct Subscription {
     pub(crate) client: Client,
 
     /// Whether the subscription is actively listening for new messages.
-    pub(crate) active: bool,
+    pub(crate) active: AtomicBool,
 }
 
 impl Subscription {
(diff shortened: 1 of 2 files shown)


spacejam:
Nice job! Much cleaner than the hack I mentioned :)

PR #86, opened 2020-07-27
Update smol dependency

stjepang:
This updates smol from v0.1 to v0.3. At the same time, we remove most of the usage of the `futures` crate.


stjepang:
Tyler gave a green light to merge, so merging.

PR #93, opened 2020-08-17
Add NKey authentication

stjepang:
This PR adds a new authentication method: `Options::with_nkey()`.

### DIFF
--- a/src/options.rs
+++ b/src/options.rs
@@ -144,33 +144,32 @@ impl Options {
         }
     }
 
-    /// Authenticate with NATS using a `.nk` file.
+    /// Authenticate with NATS using a public key and a signature function.
     ///
     /// # Example
     /// ```no_run
-    /// # fn main() -> std::io::Result<()> {
-    /// let nc = nats::Options::with_nkey("./path/to/my.nk")
+    /// let public_key = "UAMMBNV2EYR65NYZZ7IAK5SIR5ODNTTERJOBOF4KJLMWI45YOXOSWULM";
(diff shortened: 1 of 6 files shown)


src/options.rs
```diff
@@ -127,11 +127,49 @@ impl Options {
...
+            ..Default::default()
+        }
+    }
+
+    /// Authenticate with NATS using a `.nk` file.
+    ///
+    /// # Example
+    /// ```no_run
+    /// # fn main() -> std::io::Result<()> {
+    /// let nc = nats::Options::with_nkey("./path/to/my.nk")
```
derekcollison, line 152: I think nkeys are usually bundled in or env variables, want to make sure we support that model. Having nkey in a separate file may be the exception.

stjepang: How about the following two models?

* `Options::with_nkey_file("./path/to/my.nk")` - loads the seed key from a file.
* `Options::with_nkey("publickey", || load_seed())` - is given the public key and a function that loads the seed key when needed.

The Go client has `func Nkey(pubKey string, sigCB SignatureHandler) Option` but I don't know why a function that signs the nonce (`SignatureHandler`) is preferable over a function that simply loads the seed key (because we know how to sign the nonce once we have the seed key).

Which signature of these two do you like more?

1. `Options::with_nkey("publickey", || load_seed())`
2. `Options::with_nkey("publickey", |nonce| sign(nonce))`

derekcollison: Its a function because with HSM there will be no private key, you have to call out to have it sign for you etc.

stjepang: Okay, I changed `Options::with_nkey()` to take a closure that signs the nonce.

PR #97, opened 2020-09-04
Add the ability to further customize TLS connections

davidMcneil:
Previously, the only control over the TLS connection was
the ability to add new certificates with `Options::add_root_certificte`.
There are other config the user may be interested in setting. A new
method `Options::tls_config` was added to allow the user precise
control over the client's TLS configuration. `Options::native_certs`
was also added to allow a way to opt out of including the platforms
native certificate store.

Things to consider:
* `rustls::ClientConfig` is now part of `nats.rs` public interface
* I thought about not adding the `Options::native_certs` method and simply ignoring the platforms certificate store in the case that no `tls_config` was set. I thought that might be a bit surprising. Do you guys have a preference on the api?

Signed-off-by: David McNeil <mcneil.david2@gmail.com>

### DIFF
--- a/src/asynk/connector.rs
+++ b/src/asynk/connector.rs
@@ -9,7 +9,6 @@ use std::time::Duration;
 
 use async_tls::TlsConnector;
 use rand::{seq::SliceRandom, thread_rng, Rng};
-use rustls::ClientConfig;
 use smol::{io::BufReader, prelude::*, Async, Timer};
 
 use crate::asynk::proto::{self, ClientOp, ServerOp};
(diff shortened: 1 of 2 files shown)


stjepang:
@davidMcneil Thank you for the pull request!

My feeling is that we should either remove `Options::tls_config()` or remove `Options::add_root_certificate()` and `Options::native_certs()`. Those two kinds of options override each other so it's non-obvious what happens if we use all of them at the same time.

If we choose to keep `Options::tls_config()`, then we're tied a particular `rustls` version and need to catch up whenever a new version is published. If we choose to do away with it, then we aren't but TLS configuration options are limited.

I think my preference is to remove `Options::tls_config()` if `Options::add_root_certificate()` and `Options::native_certs()` are enough. But if those options are not enough for you, then let's only keep `Options::tls_config()`. What do you think?

spacejam:
I think we should avoid reexporting 3rd party crates at all costs, as it imposes significant friction over time as things start to mismatch. I think we're going to avoid adding this complexity because it's possible to add the native certs manually through the `add_root_certificate` method.

PR #98, opened 2020-09-07
Add respond on asynk::Message

### LINKED ISSUE #96: respond() not implemented for asynk::Message
Just noticed it's there for `nats::Message` but not `nats::asynk::Message`.

vbfox:
This is the same function that exists on the sync part but on the async one

Fixes #96


stjepang (approved):
Thanks!

PR #100, opened 2020-09-09
Update smol

stjepang:
This PR updates `smol` to v1.0 and replaces `serde` with `json`.

Compilation times are cut roughly in half on my 8-core machine, which means a debug build now takes 27 seconds. There are still ways to go, but it seems that the largest part of compilation is related to crypto stuff (`nkeys` and `async-tls`), which will be difficult to trim down. `smol` itself takes up only around 5 seconds.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -46,11 +46,12 @@ rustls-native-certs = "0.4.0"
 [dev-dependencies]
 criterion = "0.3.3"
 env_logger = "0.7.1"
-quicli = "0.4.0"
-structopt = "0.3.15"
 nats_test_server = { path = "nats_test_server" }
+quicli = "0.4.0"
 serde = { version = "1.0.114", features = ["derive"] }
 serde_json = "1.0.56"
+smol = "1.0.0"
+structopt = "0.3.15"
 
 [[bench]]
 name = "nats_bench"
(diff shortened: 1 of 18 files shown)


Cargo.toml
```diff
@@ -23,21 +23,19 @@ panic = 'abort'
 
 [dependencies]
 async-dup = "1.2.1"
-async-tls = "0.8.0"
+async-tls = { path = "../async-tls", default-features = false, features = ["client"] }
```
derekcollison: Why using a local copy?

Cargo.toml
```diff
@@ -23,7 +23,7 @@ panic = 'abort'
 
 [dependencies]
 async-dup = "1.2.1"
-async-tls = { path = "../async-tls", default-features = false, features = ["client"] }
+async-tls = { version = "0.10.0", default-features = false, features = ["client"] }
```
derekcollison, line 31: nvm ;)

derekcollison (approved):
In general LGTM. Love the faster compile times!

Cargo.toml
stjepang: Not anymore, that was from a previous commit. I had to temporarily use my local patched version of `async-tls` until a new version of it was published today. :)

PR #102, opened 2020-09-15
Remove crossbeam-channel from the public API

stjepang:
We're not using `crossbeam-channel` anywhere, and yet there were some remnants of it our public API. I'm removing those and also removing `const` annotations on some functions that are not really useful (they may only be useful on *constructors*, perhaps).

I'm planning to further clean up the API after the next release.


stjepang:
Merging since this should be a pretty non-controversial PR.

PR #107, opened 2020-10-07
Attempt to fix TLS

stjepang:
This is just a test right now - I changed `async-tls` to `async-rustls`.


stjepang:
Wow that worked. The build is green again!

Okay, there's definitely a bug in `async-tls` and `async-rustls` is just a fork of `tokio-rustls` that uses the standard `futures::io::{AsyncRead,AsyncWrite}` traits. I'll publish `async-rustls` on crates.io and then we can switch to it.

stjepang:
Finally, all good! Merging.

PR #108, opened 2020-10-10
Implement missing options

tiagolobocastro:
disconnect_callback
reconnect_callback
reconnect_delay_callback
max_reconnects
close_callback

[WIP]

I'm a bit confused about what max_reconnects is meant to be:
> /// Set the maximum number of reconnect attempts.
> /// If no servers remain that are under this threshold,
> /// all servers will still be attempted.
Isn't this conflicting, saying if no server remains under the max attempts, then connect anyway... ?

I'm not sure how to implement the close_callback for the drain/close situation because the client dispatch where we wait for op's does not seem to have "visibility" over the close.

### DIFF
--- a/src/asynk/client.rs
+++ b/src/asynk/client.rs
@@ -96,8 +96,9 @@ impl Client {
             unsubscribed: Arc::new(Mutex::new(HashSet::new())),
         };
 
+        let options = Arc::new(options);
         // Connector for creating the initial connection and reconnecting when it is broken.
-        let connector = Connector::new(url, options)?;
+        let connector = Connector::new(url, options.clone())?;
 
         // Spawn the client thread responsible for:
         // - Maintaining a connection to the server and reconnecting when it is broken.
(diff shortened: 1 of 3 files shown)


tiagolobocastro:
It seems that the go client attempts the max_reconnects on each server - it removes servers from its list if thei reconnects exceeds the max_reconnects: https://github.com/nats-io/nats.go/blob/efb1e55200cacd2d2bc89ac2153b94a8288fee0b/nats.go#L1209
It also resets the max_reconnects when it connects successfully.

(I've also noticed that the go client handlers have the client as the callback argument which could potentially allow for more interesting callbacks: https://github.com/nats-io/nats.go/blob/efb1e55200cacd2d2bc89ac2153b94a8288fee0b/examples/nats-sub/main.go#L108.)

So I've started adding the max_reconnects but I've noticed that failing the reconnect does seem to close the subscriptions which is less than ideal as they're left waiting on nothing. I've tried to solve this by closing the client when the connect gives up though not sure if this is appropriate.

tiagolobocastro:
I've changed the max_reconnects to match what the nats go client seems to do.

tiagolobocastro:
@stjepang I think I got all the missing options now, would you mind review this when you get a chance? Thanks

stjepang (approved):
This looks good, thank you! I just resolved merge conflicts and moved the `backoff` function into the `options.rs` file.

PR #113, opened 2020-10-19
Fix connection leak

stjepang:
This problem was reported by @dontlaugh on Slack (#rust channel),

The issue is that dropping `Connection` does not trigger thread shutdown, so if you drop it without calling `.close().await` on it, the thread simply keeps running and therefore keeps holding an active NATS connection.

The solution is to signal to the thread that it must shutdown when `Connection` is dropped.


derekcollison (approved):
LGTM

PR #114, opened 2020-10-20
Update dependencies

### LINKED ISSUE #111: Upgrade dependency to nkeys 0.0.11
The `ed25519` crate along with `signatory` have made some very large, very breaking changes to their public APIs. Any crates that use this and also use NATS can no longer use NATS until the dependency is upgraded.

stjepang:
Closes #111

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,20 +29,20 @@ async-lock = "2.3.0"
 async-net = "1.2.0"
 async-rustls = "0.1.0"
 base64-url = "1.4.5"
+fastrand = "1.4.0"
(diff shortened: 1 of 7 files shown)


derekcollison (approved):
LGTM - Check the clippy complaint though.

stjepang:
Clippy can be *really* pendantic sometimes :) In this case, it suggests using the `matches!()` macro, which is only available from Rust 1.42.0 and newer (but our minimum required Rust version is 1.41.0). I've added a rule to ignore this suggestion.

PR #117, opened 2020-10-21
Flush before dropping the connection

stjepang:
Discussed this with @derekcollison - I'm proposing a simple solution here. Changes:

1. When `nats::Connection` is dropped, we flush (and wait for the flush to complete).
2. When `nats::asynk::Connection` is dropped, we let the client thread asynchronously flush (but we don't wait for it).

cc @andrewbanchich

### DIFF
--- a/src/asynk/client.rs
+++ b/src/asynk/client.rs
@@ -128,13 +128,25 @@ impl Client {
                     });
 
                     // Spawn the main task that processes messages from the server.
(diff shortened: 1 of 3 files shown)


derekcollison (approved):
LGTM

derekcollison:
Good catch, let's make sure we have good tests that show the messages are getting out etc..

stjepang:
Unit tests have been added.

PR #119, opened 2020-11-02
Shutdown when the last reference to Connection is dropped

### LINKED ISSUE #118: asynk::Connection is shut down when a clone is dropped.

`version = "0.8.1"`

#### rustc version (`rustc --version` - we support Rust 1.37 and up)

`rustc 1.46.0 (04488afe3 2020-08-24)`

```rust
#[tokio::test]
async fn test_drop_clone() {
    let nats1 = nats::asynk::connect("nats://stan:4222").await.unwrap();
    let nats2 = nats1.clone();
    nats1.publish("foo", b"bar").await.unwrap();
    nats2.publish("foo", b"bar").await.unwrap();
    drop(nats1);
    nats2.publish("foo", b"bar").await.unwrap();
}
```

#### Expected result:

Connection should not shutdown when a clone is dropped.

#### Actual result:

Because the Drop handler of `nats1` is called shutdown is initiated but `nats2` is still around.

stjepang:
Fixes #118

### DIFF
--- a/src/asynk/connection.rs
+++ b/src/asynk/connection.rs
@@ -1,5 +1,6 @@
 use std::io::{self, Error, ErrorKind};
 use std::net::IpAddr;
+use std::sync::Arc;
 use std::time::{Duration, Instant};
 
 use crate::asynk::client::Client;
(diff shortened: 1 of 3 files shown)


stjepang:
This is a pretty simple fix - merging.

PR #120, opened 2020-11-06
Configuring client cert and private key

### LINKED ISSUE #104: mTLS / add_root_certificate clarification
I'm configuring NATS to [use mTLS](https://docs.nats.io/nats-server/configuration/securing_nats/auth_intro/tls_mutual_auth), and using [`add_root_certificate`](https://docs.rs/nats/0.8.0/nats/struct.Options.html#method.add_root_certificate) to pass in the `ca.pem` file that is just the public key of the CA.

The docs for `add_root_certificate` don't go into detail on what should be included in this file, and I don't see any equivalent to add client certs like the [Go client has](https://github.com/nats-io/nats.go#tls). I would have thought it would just be the root CA's public key, but the wording in the docs suggests it should include multiple keys.

Am I able to use mTLS with nats.rs?

stjepang:
This PR adds new method `Options::client_cert()` that configures client certificate and private key.

I also replaced `webpki` with `rustls-native-certs` so that instead of loading a static set of root certificates, we load certificates installed on the system. While playing with authentication options I was expecting `nats` to recognize my locally installed certificates by itself, which was previously not the case.

Closes #104


derekcollison (approved):
LGTM

PR #127, opened 2020-11-20
Test server changes to improve usage in tests

platy:
* Added readme to explain usage
* Wildcard support for subscriptions
* RAII to automatically shutdown server after use
* Default to use of automatic port assignment

### DIFF
--- a/nats_test_server/src/lib.rs
+++ b/nats_test_server/src/lib.rs
@@ -1,11 +1,17 @@
 use std::{
+    any::Any,
     collections::{HashMap, HashSet},
+    fmt::Display,
     io::{Read, Write},
-    net::{TcpListener, TcpStream},
+    mem::ManuallyDrop,
+    net::SocketAddr,
+    net::{TcpListener, TcpStream, ToSocketAddrs},
     sync::{
         atomic::{AtomicBool, Ordering},
-        Arc, Barrier,
+        Arc,
     },
+    thread,
+    thread::JoinHandle,
     time::{Duration, Instant},
 };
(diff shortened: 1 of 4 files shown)


spacejam:
we debugged this together and found that `localhost` is using ipv6 on the CI server and encounters issues when the client attempts to reconnect

nats_test_server/src/lib.rs
```diff
@@ -115,221 +121,364 @@ fn read_line(stream: &mut TcpStream) -> Option<String> {
...
-        if shutdown.load(Ordering::Acquire) {
-            return;
+    bugginess: Option<u32>,
+}
+
+/// A NATS test server, will be stopped on drop
+impl NatsTestServer {
+    pub fn build() -> NatsTestServerBuilder<&'static str> {
+        NatsTestServerBuilder {
+            baddr: "localhost:0",
```
spacejam, line 140:
```suggestion
            baddr: "127.0.0.1:0",
```

tests/reconnection.rs
```diff
@@ -30,31 +29,12 @@ fn reconnect_test() {
...
-                shutdown,
-                restart,
-                bugginess,
-                hop_ports,
-            )
-        }
-    });
-
-    barrier.wait();
+    let server = NatsTestServer::build().address("localhost:22222").bugginess(200).spawn();
```
spacejam, line 32:
```suggestion
    let server = NatsTestServer::build().address("127.0.0.1:22222").bugginess(200).spawn();
```

PR #129, opened 2020-11-24
Make test server accept non-blocking

platy:
Being in blocking mode initially meant that a server with no clients at the end of a test would be stuck waiting for clients and wouldn't read the shutdown flag - meaning the test couldn't complete.

This makes the `TCPListener` always non-blocking.


spacejam:
LGTM, nice work!

PR #130, opened 2020-12-01
Synchronous client implementation

stjepang:
Previously, the core client implementation was written in async style, and then the sync API was layered on top of it. Async Rust has non-negligible costs when it comes to compilation times and code maintenance. Given how much the async library ecosystem is in still in flux (and for how long it has been), maintaining crates outside the scope of this NATS client and constantly upgrading dependencies is becoming untenable.

This PR simplifies the core client implementation so that it's written in traditional sync style. The async API is presently gone but I will soon add it in a separate PR - it'll just be an adapter on top of the sync client. This way users of sync Rust don't have to bring in numerous async dependencies (improving compilation times) and code maintenance will be simpler.


spacejam:
This makes me so happy

tests/reconnection.rs
```diff
@@ -45,7 +45,7 @@ fn reconnect_test() {
         let success = success.clone();
         let shutdown = shutdown.clone();
         move || {
-            const EXPECTED_SUCCESSES: usize = 25;
+            const EXPECTED_SUCCESSES: usize = 1;
```
spacejam, line 48:
```suggestion
            const EXPECTED_SUCCESSES: usize = 25;
```

stjepang:
The tests are now passing. Windows has been re-enabled in CI. I've squashed commits and - merging!

PR #132, opened 2021-01-06
Add async-nats crate

stjepang:
### Background

The NATS client offers two kinds of APIs: blocking and asynchronous. The latest version of `nats` published on crates.io has an implementation written in async style, and the blocking API simply wraps the async API with `block_on(...)` calls.

We experienced difficulties maintaining that codebase due to instability of async Rust, general friction of writing code in async style, and hidden performance costs. Then we removed async code from the guts of the client. The current master branch contains a client written in good old sync style with a blocking API. However, the async API is missing.

### Implementation

This PR brings the async API back, which is now just a wrapper around the sync API. We use `unblock(...)` from the `blocking` crate to make functions asynchronous. Note that `unblock()` "cheats" by running code on a separate thread pool (think of it as syscall threads in the Go runtime), which is okay for functions that aren't performance sensitive (e.g. `connect()`).

However, functions like `publish()` and `next()` must be faster than that. Async `publish()` attempts to write the whole message into the buffered TCP/TLS stream in a non-blocking way. If the message doesn't fit in the buffer, we use `unblock()`. Async `next()` will first try to get the next buffered message without blocking or fall back to `unblock()`. Note that this can be optimized further in follow-up PRs, but benchmark already show performance comparable to the previous pure async implementation.

### Crate organization

@spacejam and I figured it'd be nicer to have two crates: `nats` and `async-nats` that look exactly the same except one is sync and the other is async. For example, `nats::connect(url)` becomes `async_nats::connect(url).await` when you write in async style. This PR introduces the new crate `async-nats`.

Previously we had the async API inside the `nats::asynk` module. That worked fine but the API organization wasn't as clean: `nats::connect(url)` would map to `nats::asynk::connect(url).await` but then we also had `nats::Options::new().connect(url)` that maps `nats::Options::new().connect_async(url).await`. That API didn't feel very consistent.

### Benchmarks

On my machine:

| Client | Command | Mb/sec |
|--------|---------|--------|
| Go (multithreaded) | `nats-bench -n 10000000 test` | 840 |
| Go (singlethreaded) | `GOMAXPROCS=1 nats-bench -n 10000000 test` | 990 |
| Rust (old sync) | `cargo run --release --example nats_bench test -n 10000000` | 550 |
| Rust (new sync) | `cargo run --release --example nats_bench test -n 10000000` | 900 |
| Rust (new async) | `cargo run --release --example async_nats_bench test -n 10000000` | 560 |

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -64,3 +64,9 @@ path = "examples/nats-box/main.rs"
 [[example]]
 name = "serde-json"
 path = "examples/serde-json/main.rs"
+
+[workspace]
+members = [
+    ".",
+    "async-nats",
+]
--- a/async-nats/Cargo.toml
+++ b/async-nats/Cargo.toml
@@ -1,0 +1,25 @@
+[package]
+name = "async-nats"
+version = "0.0.0"
+description = "An async Rust NATS client"
+authors = ["Derek Collison <derek@nats.io>", "Tyler Neely <tyler@nats.io>", "Stjepan Glavina <stjepan@nats.io>"]
+edition = "2018"
+license = "Apache-2.0"
+documentation = "https://docs.rs/async-nats"
+homepage = "https://github.com/nats-io/nats.rs"
+repository = "https://github.com/nats-io/nats.rs"
+readme = "README.md"
+keywords = ["nats", "client", "messaging", "api", "async"]
+categories = ["network-programming", "api-bindings"]
+
+[badges]
+maintenance = { status = "actively-developed" }
+
+[dependencies]
+blocking = "1.0.2"
+nats = { path = ".." }
+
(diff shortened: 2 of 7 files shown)


spacejam:
Looks great :)

spacejam (approved):

wdv4758h:
I install `async-nats` today, just wanna say thank you :)

jszwedko:
I stumbled upon this while upgrading nats. I was able to drop in the new crate pretty easily, but found that `async_nats::Subscription` doesn't implement `Stream` while `nats::asynk::Subscription` did. I was just curious if this was intentional or perhaps overlooked. Otherwise 👍 to the separation!

Issue #85, opened 2020-07-26
Latency Compared to Go client


dennybritz:
I've been trying to benchmark NATS for my use case. Using the [latency-tests](https://github.com/nats-io/latency-tests) tool I get something like this:

```
./latency -tr 1000 -tt 5s -sz 512
==============================
Pub Server RTT : 98µs
Sub Server RTT : 80µs
Message Payload: 512B
Target Duration: 5s
Target Msgs/Sec: 1000
Target Band/Sec: 1000K
==============================
HDR Percentiles:
10:       88µs
50:       108µs
75:       122µs
90:       137µs
99:       202µs
99.9:     253µs
99.99:    265µs
99.999:   982µs
99.9999:  982µs
99.99999: 982µs
100:      982µs
==============================
```

However, when using the Rust client and doing a very simple benchmark, my latency is ~2-5x of that. The following code is the simplest variation I have, but I tried various things: Reading and writing in a different thread, using tokio, etc, to replicate more closely what the Go benchmark does. I also tried putting explicit flush calls. The results are always the same. Am I doing something fundamentally wrong in the benchmark?

For reference, the `rtt` function gives the latency I expect, around 80 micros.

```rust
    let nc1 = nats::connect("localhost:4222").unwrap();
    let nc2 = nats::connect("localhost:4222").unwrap();
    let sub = nc2.subscribe("foo").unwrap();
    loop {
        let now = std::time::Instant::now();
        nc1.publish("foo", "").unwrap();
        sub.next().unwrap();
        println!("RTT: {:?}", now.elapsed());
        std::thread::sleep(std::time::Duration::from_millis(100));     
    }
```

```
RTT: 339.942µs
RTT: 337.414µs
RTT: 367.48µs
RTT: 357.508µs
RTT: 333.453µs
RTT: 392.305µs
RTT: 302.826µs
RTT: 495.053µs
RTT: 213.117µs
RTT: 187.48µs
RTT: 351.434µs
RTT: 419.223µs
RTT: 437.84µs
RTT: 376.603µs
RTT: 295.522µs
RTT: 303.573µs
RTT: 361.104µs
RTT: 306.155µs
RTT: 405.807µs
RTT: 313.317µs
RTT: 403.848µs
RTT: 315.56µs
RTT: 386.924µs
RTT: 379.003µs
RTT: 378.367µs
RTT: 331.52µs
RTT: 444.34µs
RTT: 354.675µs
RTT: 346.96µs
[...]

PR #141, opened 2021-01-23
set TCP_NODELAY

andy-k:


spacejam:
@andy-k have you measured this to improve long tail latencies on a real workload? Many nats workloads are throughput oriented and this just adds some complexity that people who tune TCP at the system level may not anticipate.

spacejam (changes requested):
have you measured this additional complexity to improve a real workload?

andy-k:
Thanks for the review.
It's unclear what change is requested.

I'm not familiar with how to tune TCP at the system level to enable TCP_NODELAY on all sockets.

I now understand from your comments that real nats workloads are throughput oriented, and my latency-sensitive workload is therefore not a real workload.

I had a quick look at some other libraries linked from https://nats.io/download/, and found that they are tuned to handle non-real workloads such as mine:
- https://github.com/nats-io/nats.go doesn't explicitly set TCP_NODELAY, but that's because [the default is true in Go](https://golang.org/pkg/net/#TCPConn.SetNoDelay).
- https://github.com/nats-io/nats.c/blob/cadc5e4827a838df05f69392a60ff4f7cc5b37f4/src/comsock.c#L58-L60
- https://github.com/nats-io/nats.cr/blob/aa4979137c5d98fa56baeaae05bd2a1837a46045/src/nats/connection.cr#L78
- https://github.com/nats-io/nats.js/blob/c65bd66c52c6c0d4a1b15456df46e59e230ed811/lib/nats.js#L940
- https://github.com/nats-io/nats-pure.rb/blob/0116743f866b7be90bd729596bf4034d3339a01c/lib/nats/io/client.rb#L1438
- https://github.com/DawnAngel/lua-nats/blob/980da7ac8f4c15541d114bf0cd17beff2666d9b7/src/nats.lua#L35
- https://github.com/Gr1N/nats-python/blob/d3601174b72a5dcdafc2bbb7b0bdeea775e79e5e/pynats/client.py#L168

However, I'd prefer not to have to rewrite my application into a different language.

Is there a workaround?

spacejam:
It seems that this issue relates to some sort of desired performance characteristic. How does this change impact the workloads you have in mind?

spacejam:
in #85 @saabye-io gathered some compelling metrics showing that this is likely to show a strong positive impact. Now that we have actual evidence that this change could improve things, I'm comfortable merging it

spacejam:
0.9.3 is now published which includes this change. I would be interested in seeing any additional metrics that demonstrate its impacts for various workloads.

PR #142, opened 2021-01-24
Fix typo in README: async is a wrapper around sync

sindreij:
The README claims that the sync interface is a thin wrapper around the async interface, but as far as I can tell the inverse is true.


stjepang (approved):
Thanks!

PR #144, opened 2021-01-26
Add support for multiple clients in test server

platy:
The test server now maintains separate subscriptions for different clients and forwards messages based on those.

I've also changed the ping timeout in the server from 50 micros to 50 millis


spacejam:
Looks good, thanks @platy :)

PR #145, opened 2021-01-26
Perform fault injection on the DNS lookup for servers

spacejam:


spacejam:
based on a conversation with @platy. no discovered bugs yet though.

stjepang (approved):

PR #150, opened 2021-02-09
Fix the no_echo setting

### LINKED ISSUE #147: The `no_echo` option does not actually appear to prevent echoes

```rust
use async_nats::Options;

#[tokio::main]
async fn main() {
	let connection = Options::new()
		.no_echo()
		.connect("nats://127.0.0.1:4222")
		.await
		.unwrap();	
	let subscription = connection.subscribe("test").await.unwrap();
	connection.publish("test", "no echo eh?").await.unwrap();
	assert_eq!(
		subscription.next().await.unwrap().data, 
		b"no echo eh?".to_vec()
	);
}
```

#### Expected result:

With a nats bus running at `127.0.0.1:4222` this code should hang forever as with `no_echo` set no messages should be received.

#### Actual result:

This code does not crash demonstrating a message with the same payload as the outgoing message is received despite the presence of `no_echo`.

stjepang:
Just a really silly bug... The "echo" option was specified only when set to true, but it should've been specified when it's false because true is the default. Now we always specify it in the NATS protocol.

Fixes #147

### DIFF
--- a/src/connect.rs
+++ b/src/connect.rs
@@ -57,6 +57,7 @@ impl ConnectInfo {
         let mut obj = json::object! {
             verbose: self.verbose,
             pedantic: self.pedantic,
+            echo: self.echo,
             lang: self.lang.clone(),
             version: self.version.clone(),
             tls_required: self.tls_required,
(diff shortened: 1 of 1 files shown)


spacejam:
LGTM, I'll cut a release

PR #162, opened 2021-04-01
Give access to Subscription's crossbeam channel

marjakm:
Gives access to the a crossbeam Receiver in the Subscription struct. Useful for [crossbeam_channel::select][0] macro. Diff should speak for itself. I didn't find a way to select on multiple Subscriptions in one thread without access to the crossbeam channel. Is there a better way? 

Offtopic: I used the async-nats with tokio at first, but I think [async_nats::Subsciption::next][1] implementation will lose messages if canceled by [tokio::select][2].

[0]: https://docs.rs/crossbeam-channel/0.5.0/crossbeam_channel/macro.select.html
[1]: https://github.com/nats-io/nats.rs/blob/master/async-nats/src/lib.rs#L285
[2]: https://docs.rs/tokio/0.2.18/tokio/macro.select.html

### DIFF
--- a/src/subscription.rs
+++ b/src/subscription.rs
@@ -50,6 +50,34 @@ impl Subscription {
         }))
     }
 
+    /// Get a crossbeam Receiver for subscription messages.
+    /// Useful for crossbeam_channel::select macro
+    ///
+    /// # Example
+    /// ```
+    /// # fn main() -> std::io::Result<()> {
+    /// # let nc = nats::connect("demo.nats.io")?;
+    /// # let sub1 = nc.subscribe("foo")?;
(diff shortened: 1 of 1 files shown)


spacejam:
I'm OK with merging this once the doctest is fixed. Do you need a hand with fixing the compilation error in the doctest?

marjakm:
Should be ok now. I haven't used doctests or clippy, so didn't know how to run them.

spacejam:
LGTM, thank you! I'll cut a release that includes this in a few minutes

spacejam:
0.9.10 has been published which includes this change

Issue #166, opened 2021-04-12
HMSG parsing fails when a header value contains a colon


rutgerbrf:
Make sure that these boxes are checked before submitting your issue -- thank you!

 - [x] Included below version and environment information
 - [x] Included a [Minimal, Complete, and Verifiable example](https://stackoverflow.com/help/mcve)

#### NATS version (`grep 'name = "nats"' Cargo.lock -A 1`)

nats.rs commit 357cfb62
NATS Server version 2.2.1
NATS CLI version 0.0.22

#### rustc version (`rustc --version` - we support Rust 1.41 and up)

rustc 1.51.0 (2fd73fabe 2021-03-23)

#### OS/Container environment:

Manjaro 21.0.1 Ornara

#### Steps or code to reproduce the issue:

1. Start the NATS server (localhost)
2. Clone the nats.go library (b8530c78)
3. Modify the nats-sub example in the nats.go library to print headers:
    ```diff
	diff --git a/examples/nats-sub/main.go b/examples/nats-sub/main.go
	index fc3b91e..cd3f906 100644
	--- a/examples/nats-sub/main.go
	+++ b/examples/nats-sub/main.go
	@@ -14,6 +14,7 @@
	 package main
	 
	 import (
	+	"bytes"
	 	"flag"
	 	"log"
	 	"os"
	@@ -38,7 +39,9 @@ func showUsageAndExit(exitcode int) {
	 }
	 
	 func printMsg(m *nats.Msg, i int) {
	-	log.Printf("[#%d] Received on [%s]: '%s'", i, m.Subject, string(m.Data))
	+	var hbuf bytes.Buffer
	+	m.Header.Write(&hbuf)
	+	log.Printf("[#%d] Received on [%s]: '%s', headers: %s", i, m.Subject, string(m.Data), hbuf.String())
	 }
	 
	 func main() {

    ```
4. Run the nats.go nats-sub example and subscribe to a topic that you'll use (I'm using `test`):
    ```
    $ go run main.go -s localhost test
    Listening on [test]
    ```
5. Run the nats.rs nats-box example and subscribe to the same topic:
    ```
    $ cargo run --example nats-box -- --server localhost sub test
        Finished dev [unoptimized + debuginfo] target(s) in 0.08s
         Running `target/debug/examples/nats-box --server localhost sub test`
    Listening on 'test'
    ```
6. Run the following command:
    ```
    $ nats pub -H "Test: works" test "test"
    ```
[...]

PR #168, opened 2021-04-12
Don't split a header into more than two parts

rutgerbrf:
Fixes issue #166.


spacejam:
Excellent work! Thank you very much for the clear write-up, and for providing a PR to fix it. I'll cut a release immediately that includes your fix.

spacejam:
0.9.11 has been released which includes this fix.

PR #171, opened 2021-04-19
Use unpacked split-debuginfo for speeding up macos builds of the crate.

spacejam:
context: https://jakedeichert.com/blog/reducing-rust-incremental-compilation-times-on-macos-by-70-percent/


derekcollison (approved):
LGTM

PR #176, opened 2021-04-26
Use multiple connections in the nats_bench example

spacejam:


derekcollison (approved):
In general LGTM but I still think we need to hunt down read/write imbalance issues.

spacejam:
I've solved the underlying lock-up issue in https://github.com/nats-io/nats.rs/pull/178, but I think both this and that are helpful.

PR #177, opened 2021-04-29
Add async request_timeout

erebe:
- Add support for request_timeout in async nats

Why: 
- If we get disconnected from NATS for reason, as the lib use unblock() under the hood, the application will leak threads waiting for the operation to receive a reply that will never comeback.
- Even if we get back the connection to NATS, those threads are lost in the limbo as they will never be able to receive a reply
- Wrapping `async request` within a futures::timeout, works but does not kill the underlying threads used by unblock

What:
- The PR exposes the `request_timeout` function in async-nats

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+# 0.9.15
+
+## Improvements
+
+- #177 Add support of request_timeout for async-nats
+
 # 0.9.14
 
 ## New Features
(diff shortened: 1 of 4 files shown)


spacejam:
LGTM, and thank you for updating the changelog and adding a new version! I'll get this published as soon as the tests go green (and I'll make them go green if there's a clippy issue etc...)

spacejam:
0.9.15 has been published which includes your change. Thanks again!

erebe:
Thank you, good sir :)

PR #178, opened 2021-04-29
Use fine-grained locking on read and write-related state

spacejam:
This is the real fix for the issue encountered in benchmarks where subscriptions would cause lock-ups due to slow consumers.

This comes with the introduction of a mandatory locking protocol which must always be followed to avoid deadlocks going forward. But this is important for allowing reads and writes to make progress independently of each other, and the situations that require both locks to be held are pretty infrequent.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,19 @@
-# 0.9.15
+# 0.9.16
 
 ## Improvements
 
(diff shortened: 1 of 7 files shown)


derekcollison (approved):
LGTM - Do we need to address the clippy warnings?

spacejam:
Yeah, I've just cleared those

PR #180, opened 2021-05-19
Idempotent UNSUB avoids sending multiple UNSUB messages when subscrip…

spacejam:
…tions are dropped or manually unsubscribed.


derekcollison (approved):
LGTM

Issue #179, opened 2021-05-18
Async NATS leaking subscriptions


szlend:
It's really easy to make async NATS leak subscriptions. I could be mistaken, but I believe this is caused by an `Arc` reference cycle between the IO thread pool and user threads. When the user held `Subscription` goes out of scope, the `Inner` reference in the thread pool can remain there until the next message comes in. If the topic is inactive, it can remain there indefinately.

I'm not sure what would be the best way to fix this. I think internal references and user references should probably be separate things though, so when the last user reference is released, it can cancel the blocking operations and release the inner subscription.

#### NATS version (`grep 'name = "nats"' Cargo.lock -A 1`)
0.9.16

#### rustc version (`rustc --version` - we support Rust 1.41 and up)
rustc 1.51.0

#### OS/Container environment:
Mac OS Catalina

#### Steps or code to reproduce the issue:

Here's a minimal example showcasing the issue:

```rust
use futures::StreamExt;
use std::time::Duration;

#[tokio::main]
async fn main() {
    let conn = async_nats::connect("localhost").await.unwrap();
    let task_conn = conn.clone();

    let task = tokio::spawn(async move {
        let sub = task_conn.subscribe("test").await.unwrap();

        // create a stream of subscription messages
        let mut stream = futures::stream::unfold(sub, |sub| async move {
            sub.next().await.map(|result| (result, sub))
        })
        .boxed();

        // poll the stream once to activate it
        let _ = futures::poll!(stream.next());

        // the task has to yield at some point to reproduce the issue
        tokio::task::yield_now().await;

        // the stream holding the Subscription is dropped here
    });

    // wait for the task to finish
    task.await.unwrap();
    println!("Task exited.");

    // at this point the subscription remains active
    tokio::time::sleep(Duration::from_secs(600)).await;
    drop(conn);
}
```

It's possible to workaround this issue by wrapping the `Subscription` in a struct and explicitly calling `unsubscribe` on `Drop`. This unsubscribes from the topic regardless of how many subscription `Inner` references are still active.

#### Expected result:

You can verify the number of subscriptions in NATS metrics.

Number of subscriptions: 0

#### Actual result:

Number of subscriptions: 1

spacejam:
[...]

PR #181, opened 2021-05-19
Doc-only update of async_nats::Subscription

spacejam:
documents the requirement to manually await `async_nats::Subscription::unsubscribe` due to Rust's lack of `AsyncDrop`, mentioned in https://github.com/nats-io/nats.rs/issues/179#issuecomment-843987681

### DIFF
--- a/async-nats/src/lib.rs
+++ b/async-nats/src/lib.rs
@@ -289,6 +289,11 @@ impl Connection {
 }
 
 /// A subscription to a subject.
+///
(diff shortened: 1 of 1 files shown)


derekcollison (approved):
LGTM

nazar-pc:
A question since it is very related to this. Shouldn't signature be `fn unsubscribe(self)` to indicate that subscription is no longer usable after this?

Issue #182, opened 2021-05-25
The publish method loses 2 messages after a disconnection from the NATS server


DamianoChini:
#### NATS version
name = "async-nats"
version = "0.9.16"

#### rustc version (`rustc --version` - we support Rust 1.41 and up)
rustc 1.51.0

#### OS/Container environment:
Fedora release 31 (Thirty One)

Docker for the NATS server 

#### Steps or code to reproduce the issue:
1. Create a client connection to a running NATS server
2. Stop the NATS server
3. Publish 3 messages to the NATS server
4. Restart the NATS server

#### Expected result:
I expect that after restarting the NATS server, the server receives 3 messages from the client.
Alternatively, I expect that an error is returned when the publishing fails.

#### Actual result:
The NATS server after the restart receives only the third message. The first 2 are never delivered.
The (async) NATS client returns `Ok()` for all the 3 messages, so we can not handle the messages which are not correctly published.

#### Additional details
Trying to understand when the publish fails, we also tried to `.flush().await` after every `.publish(subject, msg).await`. The flush returns an error for the first message, but for the second message it hangs until the connection with the NATS server is restored (returning `Ok()` but without delivering the message to the server). As an example please see the simplified code below.
Also the `flush_timeout()` method hangs until the connection with the NATS server is restored, without returning any error after the timeout duration expires.
```
connection.publish(subject, &msg1).await; // --> Returns Ok()
connection.flush().await; // --> Returns Err(Os { code: 32, kind: BrokenPipe, message: "Broken pipe" })

connection.publish(subject, &msg2).await; // --> Returns Ok()
connection.flush().await; // --> Hangs until connection is restored, returning Ok() but without delivering msg2

connection.publish(subject, &msg3).await; // --> Returns Ok()
connection.flush().await; // --> Hangs until connection is restored, returning Ok() and delivering msg3
```

DamianoChini:
Hello!
We saw that the first 2 messages are not lost anymore if we reset the client writer to `None` upon disconnection, in particular where the `disconnect_callback` is called [https://github.com/nats-io/nats.rs/blob/master/src/client.rs#L546](url)
```
connector.get_options().disconnect_callback.call();
self.state.write.lock().writer = None;
```
Could this solution be acceptable for you?

spacejam:
[...]

PR #183, opened 2021-05-31
Reset client writer to None upon disconnection from server

DamianoChini:
Fixes https://github.com/nats-io/nats.rs/issues/182


spacejam:
Thanks @DamianoChini ! I'll cut a new release with this update and publish it now

PR #187, opened 2021-06-16
FIFO JS consumer

spacejam:
Users generally expect that messages they receive in a stream match the order that they were published from. However, basic fault injection testing has shown that it's common in normal (non-laptop loopback) network conditions to receive messages with higher stream ID's than the next one after the last message that was consumed. While it's possible to avoid this by setting the maximum outstanding non-acked messages to 1, this slashes throughput by 2+ orders of magnitude, making this an unacceptable solution for workloads that need to process multiple messages per second. 

This solution adds a reassembly buffer that buffers messages received out-of-order. While this happens somewhat regularly, it is still unlikely to impact clusters that are experiencing nice network weather, most of the time. So, it adds a small amount of overhead in rare situations that would have caused expectations to be violated otherwise. 

This PR also adds logic to check the current stream minimum sequence ID when we receive out-of-order messages, which is necessary to adapt when the cluster starts removing messages that have either been purged or been evicted due to the configured retention policy. This does not impact the existing chance of slow consumers failing to make progress during the initialization of a quickly moving stream, but this is not exacerbated by this buffering etc... The query to the server for the lowest stream ID will only happen every 5 seconds at the most.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -190,5 +190,5 @@ jobs:
         run: |
           set -eo pipefail
           echo "msrv check"
-          rustup install 1.41.0
-          cargo +1.41.0 check
+          rustup install 1.51.0
+          cargo +1.51.0 check
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -65,7 +65,7 @@ jobs:
         with:
           go-version: '1.16'
       - name: install nats-server
-        run: go get github.com/nats-io/nats-server
+        run: go get github.com/nats-io/nats-server/v2
       - name: cargo test
         env:
           RUST_BACKTRACE: 1
@@ -112,7 +112,7 @@ jobs:
         with:
           go-version: '1.16'
       - name: install nats-server
-        run: go get github.com/nats-io/nats-server
+        run: go get github.com/nats-io/nats-server/v2
       - name: JetStream test
         env:
           RUST_LOG: trace
@@ -138,7 +138,7 @@ jobs:
         with:
           go-version: '1.16'
       - name: install nats-server
-        run: go get github.com/nats-io/nats-server
+        run: go get github.com/nats-io/nats-server/v2
       - name: JetStream test
         env:
           RUST_LOG: trace
@@ -164,7 +164,7 @@ jobs:
         with:
           go-version: '1.16'
       - name: install nats-server
-        run: go get github.com/nats-io/nats-server
+        run: go get github.com/nats-io/nats-server/v2
       - name: JetStream test
         env:
           RUST_LOG: trace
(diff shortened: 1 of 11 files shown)


derekcollison:
Only reason stream sequences are delivered out of order is when we have to redeliver a message but that is normal. If you do not want that set max delivered to 1.

Not sure we need any complex re-assembly within the client, I think that is an anti-pattern tbh.

spacejam:
@derekcollison one of the main reasons to take on the overhead of a raft-backed stream is to do things that are replicated state machine-like: databases, event sourcing, and anything else that has a strong consistency need generally has a hard requirement of processing commands in identical orders for different consumers. The overhead for client-side reassembly is pretty much unmeasurable in the steady state compared to the significant amount of effort that happens while linearizing everything with raft. But the benefits are significant, and unlock pretty much the whole reason for having taken that significant throughput hit to linearize on the server in the first place.

derekcollison:
That use case is important but only one use case of many. I worked with some folks in our ecosystem on how to do this today, the code was pretty simple TBH and we even discussed if the pattern of options and that behavior might make sense as a single option to creating a subscriber.

So for the example, we do MaxDeliver=1, AckNone, FlowControl and HBS. In the callback we track consumer sequence number since it always starts at 1 and increases by 1 regardless. If we detect a gap we create a new (ephemeral) with the stream sequence where we detected the gap. Heartbeats allow us to detect a last missed message as well.

The ordering should not be a default on client side, I may very well want them out of order and introducing head of line blocking does not make sense here.

spacejam:
As mentioned in the first comment, that works for low-throughput applications that are tolerant of a full round trip for acks before receiving any message, but this cuts throughput by 2 or 3 orders of magnitude compared to this solution. Additionally, this type of solution is required any time an `All` ack policy is to be used, otherwise there is the risk of data loss due to acking later stream ID's before earlier ones are seen. 

For pretty much any sequential data structure (queues, trees, anything), FIFO enables a low-effort replication path where API calls are translated to messages. Without FIFO, it's not eventually consistent, it's plain inconsistent, except in the rare and tricky cases where structures happen to be CRDT's. 

FIFO enables far more expressive use cases without kneecapping throughput. It seems like a no-brainer to me, and one that many users will be bitten by over time when their expectations about ordering are not true without it. Sure, it's possible to add it later, but it's simple to add it to the client, has no performance impact in the steady state, and seems to give users a more reasonable experience without much downside.

derekcollison:
The strategy I spoke about has no throughput issues at all, so maybe we are not on the same page.

spacejam:
Recording the results of the sync video chat: we will make FIFO into a configurable, with it off by default for now for parity with other clients in the ecosystem. We'll use it as a data point for getting more information about how people use it, and what the performance implications are for high-throughput situations.

caspervonb:
I'm going to respectfully close this out as stale without merge as we've adopted mostly the same interfaces laid out by the Go client for subscriptions which keep consumer details internal and went in a different direction.

PR #197, opened 2021-08-11
js config ergonomics. Closes #191.

### LINKED ISSUE #191: Improve StreamConfig ergonomics
#### Use Case:
Configure new JS streams in an easier way. In my concrete use case, I have a microservice architecture, where every service creates/updates the streams it manages (ie. the streams it publishes to, or that it listens as a work queue). At service startup, the service checks if all the streams exists, or if they need to be updated.

Currently, there are some confusing options. For instance, what is the correct way of setting no limit in `max_bytes` or `max_msgs`? Is it 0? Is it -1?  Even worse, some options use different strategies: `max_msg_size` is an `Option<i32>`, is it `None` the same as `Some(-1)`? Finally, for `no_ack` is `None` different from `Some(false)`?

I had to check the stream using the nats cli to check that I was configuring the stream correctly (which also has some gotchas, as the maximum age displays "0.00s" instead of the string "unlimited").

#### Proposed Change:
Use enums to be more expressive. Create a simple method to generate the underlying value needed by JetStream. 

For file size options (`max_bytes` and `max_msg_size`), something like that 

```rust
enum SizeLimit
{
    Unlimited,
    Bytes(u64),
   // Maybe also options for kb, mb, etc.
}    
```

For time based options (`max_age`, `duplicate_window`)

```rust
enum TimeLimit
{
    Unlimited,
    Seconds(u64),
    Minutes(u64),
    Hours(u64),
    Days(u64),
    Months(u64),
    Years(u64),
}
```

Also, subjects doesn't need to be an `Option`. A `vec![]` would mean the same. 

Other enums are possible for other options, but those are the ones I found more confusing.

Finally, all those enums and `StreamConfig` itself should implement `PartialEq` and `Eq`, making it easier to check if a Stream needs to be updated. 

I'd say that all users of the Rust client could benefit of this change, as it would make the configuration much easier to understand. Of course, this would be a breaking change, but if you guys consider this or a similar approach, doing so before 1.0.0 is the appropriate time.

#### Alternative Approaches
Improve the documentation. For every option, include the meaning of every special value (i.e, how to set a "no limits"). In some cases, it should specify the units (bytes, seconds, etc).

Another approach would be to create a `StreamConfigBuilder` type, where there could be methods like `set_max_bytes_unlimited()`  and `set_max_bytes(max: u64)`. This would give similar ergonomic improvement without the BC break.  However, the generated `StreamConfig` would still be a bit confusing itself.

spacejam:


derekcollison (approved):
LGTM

src/jetstream_types.rs
```diff
@@ -144,25 +155,35 @@ pub struct StreamConfig {
...
+    pub duplicate_window: i64,
     /// The owner of the template associated with this stream.
-    pub template_owner: Option<String>,
+    #[serde(default, skip_serializing_if = "is_default")]
+    pub template_owner: String,
+}
+
+fn is_default<T: Default + Eq>(_t: &T) -> bool {
+    // t == &T::default()
+    false
```
QuentinPerez, line 186: Just to be sure: is it wanted ?

spacejam: thanks for the catch!

PR #202, opened 2021-08-18
Add test that demonstrates the functionality of queue subscribers

spacejam:


derekcollison (approved):
LGTM - Do we have individual optional items for consumer or you always have to pass in a full consumer config?

spacejam:
@derekcollison right now you always have to pass a full consumer config, but given the `..Default::default()` syntax sugar in rust, this does not impose much API friction. We could add getters and setters for everything (enabling the builder pattern), but I think this is probably lower priority than the overall feature parity for JS and async for the time being. Especially because the partial field specification path that is enabled by the `..Default::default()` sugar very closely resembles the builder pattern as-is. But please let me know if you think it's higher priority than JS & async parity and I'll reprioritize what I'm focusing on.

PR #205, opened 2021-08-25
Add manual TLS config provision method for cases where certificates are unable to be passed-in manually. Closes #204.

### LINKED ISSUE #204: Adding CAs from string rather than file
Due to deployment concerns and a pki mess, I have my root ca compiled into my app, not in a separate file on disk.

I looked but I could not find a way to manually setup the tls config. In the go client I am able to do this via tls.Config.

spacejam:


derekcollison (approved):
LGTM

PR #209, opened 2021-08-31
auto-close async subscriptions

ccbrown:
Currently there's a somewhat hidden danger to using subscriptions: It's very easy to leak resources and occupy the `blocking` crate's thread pool indefinitely after dropping a `Future` that owns a subscription.

For example, take something like this:

```rust
async fn wait_for_thing(&self) {
    let subscription = self.connection.subscribe("things").await.unwrap();
    subscription.next().await;
    let _ = subscription.unsubscribe().await;
}
```

The crate's documentation warns that you should invoke `subscription.unsubscribe().await`, and so we do here. Unfortunately, it's very common in the async ecosystem for `Future`s to just be dropped if they're no longer needed:

```rust
async fn wait_for_thing_with_timeout(&self, timeout: Duration) {
    let _ = tokio::time::timeout(timeout, self.wait_for_thing()).await;
}
```

This sort of thing will just leave `next()` occupying the `blocking` thread pool indefinitely.

This PR makes `asynk::Subscription` clean up after itself if the user didn't call `drain` or `unsubscribe`.

### DIFF
--- a/src/asynk.rs
+++ b/src/asynk.rs
@@ -92,7 +92,10 @@ use std::fmt;
 use std::io;
 use std::net::IpAddr;
 use std::path::Path;
-use std::sync::{atomic::AtomicBool, Arc};
+use std::sync::{
+    atomic::{AtomicBool, Ordering},
+    Arc,
+};
 use std::time::Duration;
 
 use blocking::unblock;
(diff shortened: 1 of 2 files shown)


spacejam:
This is likely to cause significant reliability issues in systems that churn through async subscriptions quickly, since it causes panics at random points at runtime. `std::thread::spawn` is dangerous to use in dynamic situations because it panics when it's not able to create a new thread. In general, when creating threads, important rules to try to follow are:

1. create all necessary worker threads while starting a system, and fail to start the system if they can't all be created
2. use `try_spawn` so that it's possible to return an error to the user rather than panicking in ways that may or may not cause the process to terminate and leading to all sorts of poisoning issues (explicit and implicit).
3. for situations where dynamic thread spawning can't be avoided, it's important to always use "try_spawn" and have a correct fall-back when it fails.

Since there is not a correct fall-back when this fails, as blocking could cause unacceptable lockups of async runtimes, it should not be done. using dynamic threads 

Please continue to follow the recommended usage in the documentation.

PR #211, opened 2021-09-01
Use i64 instead of isize for representing Go's time.Duration type. Closes #208.

### LINKED ISSUE #208: ConsumerConfig::ack_wait is an isize
For a 32bits target, isize maximum value is (2^31 - 1) nanoseconds = 2.147s.

Don't know if this is something for upstream compatibility with Nats, but 2 seconds looks like a short time limit for some kind of workloads in 32 bits systems. 

An i64 (or even a u64 if possible, don't think negative values have a meaning here) is ok, as (2^63 - 1) nanoseconds > 500years.

On a related note, the standard library uses u128 for nanoseconds (https://doc.rust-lang.org/stable/std/time/struct.Duration.html#method.as_nanos). Doesn't look necessary, but if Nats supports that, this would ease the conversion if some application tries to store the configuration as a Duration and then translates it to the `ConsumerConfig` struct (in my use case, I use the same time span for timeout-like stuff, pretty much everything else requires a `Duration` instance).

name = "nats"
version = "0.13.0"

32 bits SO

N/A

#### Expected result:
N/A

#### Actual result:
N/A

spacejam:


derekcollison (approved):
LGTM

PR #230, opened 2021-10-28
fix: add read and write timeouts to the tcp stream

tiagolobocastro:
The tcp stream to the nats-server can get stuck if the server is not
shutdown gracefully. Even the connection lost callback does not get
triggered because the stream is somehow stuck.

Adding the timeouts seems to resolve this issue, at least on my test
environment.

By default the timeouts are not set so the existing behavior is kept
but they can be set to a Duration value or None which is then set on
the tcp stream.

### DIFF
--- a/src/asynk.rs
+++ b/src/asynk.rs
@@ -856,4 +856,43 @@ impl Options {
             inner: self.inner.add_root_certificate(path),
         }
     }
+
+    /// Sets the tcp read timeout to the timeout specified
+    ///
+    /// # Examples
+    /// ```no_run
+    /// # smol::block_on(async {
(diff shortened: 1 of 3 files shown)


tiagolobocastro:
Ok I see so looks like it's the client that sends pings to the server? https://docs.nats.io/developing-with-nats/connecting/pingpong
At which point then that's the proper fix to do here then, implement the ping/pong? @derekcollison

derekcollison:
Yes, part of the original todo in the README that we did not get to, but I will add it this weekend and do no responders, etc.

tiagolobocastro:
Sounds good, I think we can close this one then.

PR #232, opened 2021-11-01
Parse inline status and description in `Headers::try_from`

caspervonb:


Jarema:
@caspervonb Go client checks version line against `NATS/1.0`. I would do the same here, just for consistency.

src/headers.rs
```diff
@@ -112,8 +112,28 @@ impl TryFrom<&[u8]> for Headers {
         };
 
         if let Some(line) = lines.next() {
-            if !line.starts_with("NATS/") {
-                return parse_error("version line does not begin with NATS/");
+            let mut parts = line.splitn(3, ' ');
+
+            if let Some(v) = parts.next() {
+                if !v.starts_with("NATS") {
```
derekcollison: Rust ecosystem more about inline or making those things const? I did make them const but not sure what is idiomatic these days.

src/headers.rs
```diff
@@ -112,8 +112,28 @@ impl TryFrom<&[u8]> for Headers {
...
+                    return parse_error("version line does not begin with NATS/");
+                }
+            }
+
+            if let Some(v) = parts.next() {
```
derekcollison, line 123: Most of the time will not have inline status, so be good not to spend cycles when not needed. Hence the top level check on line length in the other PR.

src/headers.rs
```diff
@@ -112,8 +112,28 @@ impl TryFrom<&[u8]> for Headers {
         };
 
         if let Some(line) = lines.next() {
-            if !line.starts_with("NATS/") {
-                return parse_error("version line does not begin with NATS/");
+            let mut parts = line.splitn(3, ' ');
```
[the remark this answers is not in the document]
Jarema: any additional spaces will become `` parts on their own.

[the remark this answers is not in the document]
derekcollison: So that means this would break yes?

src/headers.rs
caspervonb: Constants was a good call 🙂

src/headers.rs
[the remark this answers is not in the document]
caspervonb: Yeah something like this would fail: `NATS/1.0    200      DESC` in that status and description would only be empty spaces.

src/headers.rs
caspervonb: Yeah I ended up with more or less your take in the end minus one allocation.

src/headers.rs
```diff
@@ -116,28 +118,27 @@ impl TryFrom<&[u8]> for Headers {
...
 
         if let Some(line) = lines.next() {
-            let mut parts = line.splitn(3, ' ');
-
-            if let Some(v) = parts.next() {
-                if !v.starts_with(VERSION_LINE) {
-                    return parse_error("version line does not begin with NATS/");
-                }
+            if !line.starts_with(VERSION_LINE) {
+                return parse_error("version line does not begin with NATS/1.0");
```
derekcollison, line 122: Would call it header line, follows HTTP process.

derekcollison (approved):
LGTM - Minor suggestion.

Merge after #231 imo

PR #237, opened 2021-11-01
Add support for continuation lines in `Headers::try_from`

caspervonb:


src/headers.rs
```diff
@@ -157,12 +161,19 @@ impl TryFrom<&[u8]> for Headers {
...
         };
 
-        for line in lines {
+        while let Some(line) = lines.next() {
             let splits = line.splitn(2, ':').map(str::trim).collect::<Vec<_>>();
             match splits[..] {
                 [k, v] => {
                     let entry = inner.entry(k.to_string()).or_insert_with(HashSet::default);
-                    entry.insert(v.to_string());
+
```
derekcollison, line 169: Does this skip efficiently when its just **NATS/1.0**

caspervonb: If the entire header is just `NATS/1.0` it wouldn't reach inside the loop at all as `next()` above would return `None` on the first call.

derekcollison: nvm this just cleans up the normal k/v insertion code, got it.

derekcollison (approved):
LGTM

src/headers.rs
caspervonb: Sort of, It's turned into a "peekable" iter to allow continuation lines to "take" additional lines below.

PR #239, opened 2021-11-01
Bump MSRV to 1.52.0

caspervonb:
This bumpos the minimum supported rust version to 1.52.0 which is required for `str::split_once`.


derekcollison (approved):
LGTM

PR #241, opened 2021-11-02
Allow for inline header description with spaces

caspervonb:


src/headers.rs
```diff
@@ -139,21 +139,30 @@ impl TryFrom<&[u8]> for Headers {
...
-                        .or_insert_with(HashSet::default);
-                    entry.insert(status.to_string());
-
-                    if let Some(description) = parts.next() {
-                        let entry = inner
-                            .entry(DESCRIPTION_HEADER.to_string())
-                            .or_insert_with(HashSet::default);
-                        entry.insert(description.to_string());
+            if let Some(slice) = line.get(HEADER_LINE_LEN..).map(|s| s.trim()) {
+                match slice.split_once(' ') {
```
derekcollison, line 143: What if 2 spaces between code and description?

caspervonb: Got a test case for that, splits on first space then trims 🙂

derekcollison (approved):
LGTM

PR #242, opened 2021-11-02
feat idle heartbeat

caspervonb:
- Allow for inline header description with spaces
- Add support for consumer idle heartbeats

### DIFF
--- a/tests/jetstream.rs
+++ b/tests/jetstream.rs
@@ -2,6 +2,8 @@ use std::io;
 
 mod util;
 use nats::jetstream::*;
+use std::collections::HashSet;
+use std::iter::{FromIterator, IntoIterator};
 pub use util::*;
 
 #[test]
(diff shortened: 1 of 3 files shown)


tests/jetstream.rs
```diff
@@ -165,3 +167,45 @@ fn jetstream_libdoc_test() {
...
+                deliver_subject: Some("subject".to_string()),
+                durable_name: Some("consumer".to_string()),
+                idle_heartbeat: 200000000,
+                ..Default::default()
+            },
+        )
+        .unwrap();
+
+    consumer
+        .process(|message| {
```
derekcollison, line 195: We want these HBs handled under the covers like the Go client and not exposed to the application unless the HB detects that we missed messages etc.

caspervonb:
Doing new abstractions and types for jetstream subscription, handling that there and closing this out.

PR #244, opened 2021-11-03
Allow setting a jetstream api prefix from a domain

caspervonb:
https://github.com/nats-io/nats-architecture-and-design/issues/52

### DIFF
--- a/src/options.rs
+++ b/src/options.rs
@@ -23,6 +23,74 @@ use crate::auth_utils;
 use crate::secure_wipe::SecureString;
 use crate::Connection;
 
+/// `JetStream` options
+#[allow(clippy::module_name_repetitions)]
+pub struct JetStreamOptions {
+    pub(crate) api_prefix: String,
+}
+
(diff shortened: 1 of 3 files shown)


src/options.rs
```diff
@@ -524,6 +524,16 @@ impl Options {
         self
     }
 
+    /// Set a custom `JetStream` API prefix from a domain.
+    ///
+    pub fn jetstream_api_prefix_from_domain(mut self, domain: &str) -> Self {
```
derekcollison, line 529: fn name seems really long, better way to do this?

For Go client its options to getting the JetStream context which I like.

derekcollison: I like it, would just say domain though like we do in Go client. Do APIPrefix() option and Domain() option. In Rust version of course.

caspervonb: As a first pass, let's just embed `JetStreamOptions` into `Options` (this PR).

derekcollison (approved):
LGTM

PR #252, opened 2021-11-08
Re-work JetStream consumer interface

caspervonb:
Adds support for jetstream subscriptions as described by ADR-15 and ordered consumers as described by ADR-17.

There's one defect from ADR-15 as it is right now:

> If no message or status is received within a certain time period, an alarm should be surfaced to the user in a way that makes sense in the client language, for example by notifying the user through the asynchronous error callback.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -26,7 +26,7 @@ use crossbeam_channel::RecvTimeoutError;
 use parking_lot::Mutex;
 
 use crate::connector::{Connector, NatsStream};
-use crate::message::Message;
+use crate::message::{Message, MessagePreprocessor};
 use crate::proto::{self, ClientOp, ServerOp};
 use crate::{header::HeaderMap, inject_delay, inject_io_failure, Options, ServerInfo};
 
@@ -82,6 +82,7 @@ struct Subscription {
     subject: String,
     queue_group: Option<String>,
     messages: channel::Sender<Message>,
+    preprocess: MessagePreprocessor,
 }
 
 /// A NATS client.
@@ -392,6 +393,7 @@ impl Client {
                 subject: subject.to_string(),
                 queue_group: queue_group.map(ToString::to_string),
                 messages: sender,
+                preprocess: Box::new(|_| false),
             },
         );
 
@@ -400,6 +402,57 @@ impl Client {
         drop(write);
 
         Ok((sid, receiver))
+    }
+
+    /// Resubscribes an existing subscription by unsubscribing from the old subject and subscribing
+    /// to the new subject returning a new sid while retaining the existing channel receiver.
+    pub(crate) fn resubscribe(&self, old_sid: u64, subject: &str) -> io::Result<u64> {
+        // Inject random delays when testing.
+        inject_delay();
+
+        let mut write = self.state.write.lock();
+        let mut read = self.state.read.lock();
+
+        let subscription = read
+            .subscriptions
+            .remove(&old_sid)
+            .ok_or_else(|| io::Error::new(io::ErrorKind::InvalidInput, "subscription not found"))?;
+
+        // Send an UNSUB and SUB messages.
+        if let Some(writer) = write.writer.as_mut() {
+            proto::encode(
+                writer,
+                ClientOp::Unsub {
+                    sid: old_sid,
+                    max_msgs: None,
+                },
+            )?;
+        }
+
+        // Generate a subject ID.
+        let new_sid = write.next_sid;
+        write.next_sid += 1;
+
+        let queue_group = subscription.queue_group.clone();
+        read.subscriptions.insert(new_sid, subscription);
+
+        if let Some(writer) = write.writer.as_mut() {
+            proto::encode(
+                writer,
+                ClientOp::Sub {
+                    sid: new_sid,
+                    subject,
+                    queue_group: queue_group.as_deref(),
+                },
+            )?;
+            write.flush_kicker.try_send(()).ok();
+        }
+
+        // NB see locking protocol for state.write and state.read
+        drop(read);
+        drop(write);
+
+        Ok(new_sid)
     }
 
     /// Unsubscribes from a subject.
@@ -431,6 +484,30 @@ impl Client {
         // NB see locking protocol for state.write and state.read
         drop(read);
         drop(write);
+
+        Ok(())
+    }
+
+    /// Sets the message preprocessor for a subscription.
+    ///
+    /// This predicate is called afer a message arrives and before it is queued up to subscription
+    /// channel.
+    ///
+    /// If the predicate returns true the message is considered consumed and is not queued on the
+    /// channel.
+    pub(crate) fn set_preprocessor(
+        &self,
+        sid: u64,
+        preprocess: MessagePreprocessor,
+    ) -> io::Result<()> {
+        let mut read = self.state.read.lock();
+
+        if let Some(subscription) = read.subscriptions.get_mut(&sid) {
+            subscription.preprocess = preprocess;
+        }
+
+        // NB see locking protocol for state.write and state.read
+        drop(read);
 
         Ok(())
     }
@@ -779,6 +856,13 @@ impl Client {
                             double_acked: Default::default(),
                         };
 
+                        // Preprocess and drop the message from the buffer if it the predicate
+                        // returns true
+                        let preprocess = &subscription.preprocess;
+                        if preprocess(&msg) {
+                            continue;
+                        }
+
                         // Send a message or drop it if the channel is
                         // disconnected or full.
                         subscription.messages.try_send(msg).ok();
@@ -803,6 +887,13 @@ impl Client {
                             client: Some(self.clone()),
                             double_acked: Default::default(),
                         };
+
+                        // Preprocess and drop the message from the buffer if it the predicate
+                        // returns true
+                        let preprocess = &subscription.preprocess;
+                        if preprocess(&msg) {
+                            continue;
+                        }
 
                         // Send a message or drop it if the channel is
                         // disconnected or full.
--- a/src/jetstream.rs
+++ b/src/jetstream.rs
@@ -29,7 +29,7 @@
 //! # Ok(()) }
 //! ```
 //!
-//! Add a new stream with specific options set:
+//! Create a new stream with configuration:
 //!
 //! ```no_run
 //! # fn main() -> Result<(), Box<dyn std::error::Error>> {
@@ -48,7 +48,7 @@
 //! # Ok(()) }
 //! ```
 //!
-//! Create and use a new default consumer (defaults to Pull-based, see the docs for [`ConsumerConfig`](struct.ConsumerConfig.html) for how this influences behavior)
+//! Create a new consumer:
 //!
 //! ```no_run
 //! # fn main() -> Result<(), Box<dyn std::error::Error>> {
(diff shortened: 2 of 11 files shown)


src/jetstream.rs
```diff
@@ -57,7 +57,7 @@
 //!
 //! js.add_stream("my_stream")?;
 //!
-//! let consumer: nats::jetstream::Consumer = js.add_consumer("my_stream", "my_consumer")?;
+//! let consumer: nats::jetstream::ConsumerSubscription = js.add_consumer("my_stream", "my_consumer")?;
```
derekcollison: Consumers are separate from Subscriptions. This might need to be reconsidered a bit.

derekcollison: I think Consumer is more correct here.

src/jetstream.rs
```diff
@@ -93,7 +93,7 @@
 //! let nc = nats::connect("my_server::4222")?;
 //! let js = nats::jetstream::new(nc);
 //!
-//! let consumer_res = js.existing("my_stream", "non-existent_consumer");
+//! let consumer_res = js.bind("my_stream", "non-existent_consumer");
```
derekcollison, line 96: I like short names too, but maybe bind_existing, wdyt?

src/jetstream.rs
```diff
@@ -99,7 +99,7 @@
 //! assert!(consumer_res.is_err());
 //!
 //! // this will create the consumer if it does not exist already
-//! let consumer = js.create_or_bind("my_stream", "existing_or_created_consumer")?;
+//! let consumer = js.subscribe("my_stream", "existing_or_created_consumer")?;
```
derekcollison, line 102: Subscribe should always take a subject as first arg, so if this is still trying to bind to the stream name this is mis-named.

src/jetstream.rs
```diff
@@ -57,7 +57,7 @@
 //!
 //! js.add_stream("my_stream")?;
 //!
-//! let consumer: nats::jetstream::ConsumerSubscription = js.add_consumer("my_stream", "my_consumer")?;
+//! let consumer: nats::jetstream::ConsumerSubscription = js.subscribe("my_stream", "my_consumer")?;
```
derekcollison, line 60: Same, subscribe should always take a subject and we do the work underneath the covers.

However, to create a sub from a consumer that you bind to should be possible.

```rust
let sub = consumer.sub();
```

derekcollison (dismissed):

src/jetstream.rs
caspervonb: Ended up settling on discarding `Consumer` completely.
Lets come back to a potential lower level interface once we have feature parity with the other clients.

src/jetstream.rs
```diff
@@ -1050,175 +1086,157 @@ impl JetStream {
...
 }
 
-/// `JetStream` reliable consumption functionality.
-pub struct Consumer {
-    /// The underlying NATS client
-    pub js: JetStream,
+#[derive(Debug)]
+pub(crate) enum ConsumerKind {
+    Ephemeral,
+    Persistent,
```
Jarema, line 1092: `Durable` is more NATS-idiomatic way to call it instead of Persistent.

src/jetstream_types.rs
```diff
@@ -550,23 +561,178 @@ pub struct NextRequest {
...
+    /// Configures the consumer to receive all the messages from a Stream.
+    pub fn deliver_all(mut self) -> Self {
+        self.config.deliver_policy = DeliverPolicy::All;
+        self
+    }
+
+    /// DeliverLast configures the consumer to receive messages
+    /// starting with the latest one.
+    pub fn deliver_last(mut self) -> Self {
+        self.config.deliver_policy = DeliverPolicy::All;
```
Jarema, line 617: should be `DeliverPolicy::Last`

Jarema (changes requested):

src/jetstream.rs
caspervonb: Does it have to be durable tho? Just meant to communicate internally if we bound to an existing consumer or created an ephemeral one.

caspervonb: Renamed this to a simple but still explicit `ConsumerOwnership` boolean enum (`Yes` or `No`).

caspervonb: I'll go ahead and mark this as resolved.

src/message.rs
```diff
@@ -88,6 +91,58 @@ impl Message {
...
+                if set.get("FlowControl Request").is_some() {
+                    return true;
+                }
+            }
+        }
+
+        false
+    }
+
+    // Helper for detecting flow control messages.
```
caspervonb, line 122:
```suggestion
    // Helper for detecting idle heartbeat messages.
```

src/jetstream_types.rs
```diff
@@ -98,12 +107,18 @@ pub struct ConsumerConfig {
...
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    /// Deliver group to use.
+    pub deliver_group: Option<String>,
     /// Allows for a variety of options that determine how this consumer will receive messages
     pub deliver_policy: DeliverPolicy,
     /// Used in combination with `DeliverPolicy::ByStartSeq` to only select messages arriving
     /// after this sequence number.
     #[serde(default, skip_serializing_if = "is_default")]
-    pub opt_start_seq: i64,
+    pub opt_start_seq: u64,
```
caspervonb, line 121: Not sure why this was signed.

src/jetstream_types.rs
```diff
@@ -123,7 +138,7 @@ pub struct ConsumerConfig {
     pub replay_policy: ReplayPolicy,
     /// The rate of message delivery in bits per second
     #[serde(default, skip_serializing_if = "is_default")]
-    pub rate_limit: i64,
+    pub rate_limit: u64,
```
caspervonb, line 141: Ditto, also not sure why it was unsigned.

src/jetstream_types.rs
```diff
@@ -135,6 +150,15 @@ pub struct ConsumerConfig {
...
     pub max_ack_pending: i64,
+    /// Only deliver headers without payloads.
+    #[serde(default, skip_serializing_if = "is_default")]
+    pub headers_only: bool,
+    /// Enable flow control messages
+    #[serde(default, skip_serializing_if = "is_default")]
+    pub flow_control: bool,
+    /// Enable idle heartbeat messages
+    #[serde(default, skip_serializing_if = "is_default")]
+    pub idle_heartbeat: i64,
```
caspervonb, line 161: This is signed altho it looks odd, because time.Duration in Go is signed as far as I know.

src/jetstream_types.rs
```diff
@@ -553,23 +584,214 @@ pub struct NextRequest {
...
+
+/// Options for subscription
+#[derive(Debug, Default, Serialize, Deserialize, Clone)]
+pub struct SubscribeOptions {
+    // For consumer binding:
+    pub(crate) bind_only: bool,
+    pub(crate) stream_name: Option<String>,
+    pub(crate) consumer_name: Option<String>,
+
+    // For consumer configuration:
```
caspervonb, line 605: Because defaults don't always indicate intent, configuration is duplicated here as optional values.

src/jetstream_types.rs
```diff
@@ -553,23 +584,214 @@ pub struct NextRequest {
...
+    }
+
+    /// Sets the description used for the created consumer.
+    pub fn description(mut self, description: String) -> Self {
+        self.description = Some(description);
+        self
+    }
+
+    /// Sets the durable name for the created consumer.
+    pub fn durable_name(mut self, consumer: String) -> Self {
```
caspervonb, line 669: Should durable mutually exclusive along with bind and ordered as a top level option?

src/jetstream_push_subscription.rs
```diff
@@ -0,0 +1,544 @@
...
+            .connection
+            .0
+            .client
+            .unsubscribe(self.0.sid.load(Ordering::Relaxed))?;
+
+        // Delete the consumer, if we own it.
+        if self.0.consumer_ownership == ConsumerOwnership::Yes {
+            self.0
+                .context
+                .delete_consumer(&self.0.stream, &self.0.consumer)?;
```
caspervonb, line 470: Double unsubscribing is a failure, probably should be a no-op.

src/jetstream_types.rs
Jarema: yep, Go's `time.Duration` is int64, signed.

src/jetstream_push_subscription.rs
Jarema: In Go client double unsubscribe does not error:
```Go
	// Already unsubscribed
	if s == nil {
		return nil
	}
```
So I will follow the same approach.

src/client.rs
```diff
@@ -402,6 +404,56 @@ impl Client {
...
+            .remove(&old_sid)
+            .ok_or_else(|| io::Error::new(io::ErrorKind::InvalidInput, "subscription not found"))?;
+
+        // Send an UNSUB and SUB messages.
+        if let Some(writer) = write.writer.as_mut() {
+            proto::encode(
+                writer,
+                ClientOp::Unsub {
+                    sid: old_sid,
+                    max_msgs: Some(0),
```
Jarema, line 427: probably could be just None.

src/jetstream.rs
```diff
@@ -731,33 +664,692 @@ impl JetStream {
...
+        maybe_queue: Option<&str>,
+        maybe_options: Option<&SubscribeOptions>,
+    ) -> io::Result<PushSubscription> {
+        // If no stream name is specified the subject cannot be empty.
+        if subject.is_empty()
+            && maybe_options
+                .map(|options| options.stream_name.as_ref())
+                .is_none()
+        {
+            return Err(io::Error::new(
```
Jarema, line 748: We could probably avoid that runtime error with builder.

caspervonb: You're thinking only taking in a single argument which is a builder right?

Yeah I'd want to stick with option variants so the simple "base" signature is the same as it is for NATS subscriptions, e.g `JetStream::subscribe` and `Connection::subscribe` take in the same arguments.

src/jetstream.rs
```diff
@@ -1030,7 +1569,8 @@ impl JetStream {
     where
         Res: DeserializeOwned,
     {
-        let res_msg = self.nc.request(subject, req)?;
+        let res_msg = self.connection.request(subject, req)?;
+        // println!("res: {:?}", std::str::from_utf8(&res_msg.data).unwrap());
```
Jarema, line 1573: remove :)

src/jetstream_types.rs
```diff
@@ -553,23 +584,215 @@ pub struct NextRequest {
...
+    /// Enables flow control
+    pub fn enable_flow_control(mut self) -> Self {
+        self.flow_control = Some(true);
+        self
+    }
+
+    /// Enables hearbeat messages to be sent.
+    #[allow(clippy::cast_possible_truncation)]
+    pub fn idle_heartbeat(mut self, interval: Duration) -> Self {
+        self.idle_heartbeat = Some(interval.as_nanos() as i64);
```
Jarema, line 793: Any reason to not use `Duration` as a type for `idle_hearbeat` ?

src/jetstream.rs
```diff
@@ -731,33 +664,692 @@ impl JetStream {
...
+            }
+
+            // Check for configuration mismatches.
+            if let Some(options) = maybe_options {
+                if options.durable_name.is_some()
+                    && options.durable_name != info.config.durable_name
+                {
+                    return Err(io::Error::new(
+                        io::ErrorKind::Other,
+                        format!("configuration requests durable name to be {:?}, but consumer's value is {:?}", options.durable_name, info.config.durable_name
```
Jarema, line 885: I don't like our usage of `io::Error` in general, but it shows it weaknesses especially with those parametrized error, all of kind `Other`.

This (params) will make it impossible to check for those explicit errors, though on other hand, that additional information is very valuable for the user.

src/jetstream_types.rs
caspervonb: Mainly just because the other fields are represented as integers.
So all of them have to be done for consistency and this is already quite a large change.

caspervonb: I'll sort out the serialisation issues in a single commit when re-basing.

src/jetstream.rs
```diff
@@ -731,33 +664,692 @@ impl JetStream {
...
+                ));
+            }
+
+            if let Some(deliver_group) = info.config.deliver_group.as_ref() {
+                if let Some(queue) = maybe_queue {
+                    if deliver_group != queue {
+                        return Err(io::Error::new(
+                            io::ErrorKind::Other,
+                            format!(
+                                "cannot create a queue subscription {} for a consumer with a deliver group {}",
```
Jarema, line 844: I would prefer error telling user why he cannot create the sub  "cannot create a queue subscription if deliver name is set and does not match queue name"

caspervonb: No opinion either way, error messages are taken from the Go client.

src/jetstream_push_subscription.rs
```diff
@@ -0,0 +1,544 @@
...
+            consumer_ack_policy: consumer_info.config.ack_policy,
+            consumer_ownership,
+            messages,
+            context,
+        }))
+    }
+
+    /// Filters and responds to protocol messages returning false if a message
+    /// should be filtered out.
+    fn filter(&self, message: &Message) -> bool {
```
Jarema, line 88: The naming is confusing.
I had to  check comment to understand when filter returns true.

As this is very specific filter, wouldn't specific name fit better?
i.e: `is_proto_message` ?
It would make the below if on `next()` also more read-friendly.

src/jetstream_push_subscription.rs
```diff
@@ -0,0 +1,544 @@
...
+    ///   println!("Received {}", msg);
+    /// }
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub fn try_next(&self) -> Option<Message> {
+        match self.0.messages.try_recv().ok() {
+            Some(msg) => {
+                if !self.filter(&msg) {
+                    return self.next();
```
Jarema, line 147: should use `self.try_next()`

src/jetstream_push_subscription.rs
```diff
@@ -0,0 +1,544 @@
...
+    /// })?;
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub fn process<R, F: Fn(&Message) -> io::Result<R>>(&mut self, f: F) -> io::Result<R> {
+        let next = self.next().unwrap();
+
+        let result = f(&next)?;
+        if self.0.consumer_ack_policy != AckPolicy::None {
+            next.ack().ok();
```
Jarema, line 284: I know that it's coming from old implementation of `process` but I'm not conviced on mapping `ack()` to Option and discarding the error telling the user to use `double_ack` (or `ack_sync` in our case) instead. After all, double_ack is different mechanism (using request-response), not just `ack` with error handled.

src/jetstream_push_subscription.rs
```diff
@@ -0,0 +1,544 @@
...
+    pub fn process_timeout<R, F: Fn(&Message) -> io::Result<R>>(
+        &mut self,
+        timeout: Duration,
+        f: F,
+    ) -> io::Result<R> {
+        let next = self.next_timeout(timeout)?;
+
+        let ret = f(&next)?;
+        if self.0.consumer_ack_policy != AckPolicy::None {
+            next.ack().ok();
```
Jarema, line 327: Same as above.

src/jetstream.rs
```diff
@@ -731,33 +664,692 @@ impl JetStream {
...
+
+                if options.idle_heartbeat.is_some()
+                    && options.idle_heartbeat.unwrap() != info.config.idle_heartbeat
+                {
+                    return Err(io::Error::new(
+                        io::ErrorKind::Other,
+                        format!("configuration requests heartbeat to be {:?}, but consumer's value is {:?}", options.idle_heartbeat, info.config.idle_heartbeat
+                    )));
+                }
+                // ...
```
Jarema, line 1011: cleanup :).

src/jetstream_push_subscription.rs
caspervonb: Yeah I just brought this over as it was.

Tangental to this, I wondered if we should just do `with_handler` Ala `Subscription::with_handler` and then also an `with_ack_handler` that always acknowledges.

src/jetstream_push_subscription.rs
caspervonb: Open to suggestions.
It has side effects tho, so naming shouldn't be a `is_` predicate.

derekcollison: I have not idea what this is supposed to do? Consumers can be configured to have a filtered subject, and they actually should have one in all client impl. But I have not idea what this fn would be doing?

caspervonb: Filters and handles protocol messages, e.g flow control.

src/jetstream.rs
caspervonb: Agree, but we should take a pass and strongly type all the errors everywhere for this to be useful.

src/client.rs
```diff
@@ -82,6 +82,7 @@ struct Subscription {
     subject: String,
     queue_group: Option<String>,
     messages: channel::Sender<Message>,
+    filter: MessageFilter,
```
derekcollison, line 85: What is this for? Is this just a filter subject that is attached to a consumer? Struggling to realize what this is here for.

src/jetstream.rs
```diff
@@ -734,33 +667,702 @@ impl JetStream {
...
+    /// ```
+    pub fn queue_subscribe(&self, subject: &str, queue: &str) -> io::Result<PushSubscription> {
+        self.do_push_subscribe(subject, Some(queue), None)
+    }
+
+    /// Creates a push-based consumer subscription with a queue group and options.
+    ///
+    /// If a durable name is not set within the options provided options then the queue group will
+    /// be used as the durable name.
+    ///
```
derekcollison, line 734: Should have simple example imo.

src/jetstream.rs
derekcollison: We should include the error from the server at minimum.

src/jetstream.rs
derekcollison: What would be the alternative. Our own errors yes?

derekcollison: Which ever way we go, a simple example with main and the ability to use '?' is desired.

src/jetstream_push_subscription.rs
derekcollison: ok, so this can process HBs and FCs. Got it. Might have a better name to avoid confusion.

A flow like..

```rust
if control_msg(&m) {
  process_control{}
} else {
  do_callback()
```
}

src/jetstream_push_subscription.rs
derekcollison: That might work here, since pull is always manual ack IIRC.

src/jetstream_types.rs
derekcollison: Should we use chrono's Duration for ease of use for our users?

derekcollison: Or normal time Duration even.

src/jetstream_types.rs
derekcollison: They could go over wire as i64 if that is what other side needs but feels like a higher cognitive load for user converting between nanosecs and things that make sense to them.

src/message.rs
```diff
@@ -24,6 +24,8 @@ use crate::{
     header::{self, HeaderMap},
 };
 
+/// A predicate used to filter messages.
+pub(crate) type MessageFilter = Box<dyn Fn(&Message) -> bool + Send + Sync>;
```
derekcollison, line 28: Might want to explore another name to avoid FilterSubject option/property on consumers.

src/message.rs
derekcollison: detecting control messages?

tests/jetstream.rs
```diff
@@ -247,162 +247,255 @@ fn jetstream_publish() {
...
 
-    js.add_stream("my_stream").unwrap();
-    nc.publish("my_stream", "1").unwrap();
-    nc.publish("my_stream", "2").unwrap();
-    nc.publish("my_stream", "3").unwrap();
-    nc.publish("my_stream", "4").unwrap();
+    js.add_stream(&StreamConfig {
+        name: "TEST".to_string(),
+        subjects: Some(vec![
+            "foo".to_string(),
```
derekcollison, line 424: Anyway to allow just vec!["foo", "bar"], etc?

src/client.rs
caspervonb: Filtering and processing as messages arrive on the write side of the subscription.
Used to handle control messages like heartbeats, flow control, detect sequence mismatches, etc.

src/jetstream.rs
```diff
@@ -48,139 +48,69 @@
...
+    sync::atomic::Ordering,
 };
 
+use std::sync::{Arc, Mutex};
+use std::thread;
+
 use serde::{de::DeserializeOwned, Deserialize, Serialize};
 use serde_repr::{Deserialize_repr, Serialize_repr};
 
+const ORDERED_IDLE_HEARTBEAT: i64 = 5_000_000_000;
```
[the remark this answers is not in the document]
caspervonb: `chrono::Duration` for all nanosecond durations, including existing properties like max_age and such would be preferable imo.

src/jetstream.rs
caspervonb: Server errors are always reported, this is client side validation.

tests/jetstream.rs
caspervonb: Breaking change but yes, empty vector is as meaningful as an optional one imo.

src/jetstream.rs
caspervonb: Yeah it would be to strongly type every single error that is being returned across the entire crate.
Should be considered but punting from this pull request.

src/message.rs
caspervonb: Went with message preprocessor.

src/client.rs
```diff
@@ -435,6 +487,20 @@ impl Client {
         Ok(())
     }
 
+    /// Sets the message filter for a subscription.
+    pub(crate) fn set_filter(&self, sid: u64, message_filter: MessageFilter) -> io::Result<()> {
```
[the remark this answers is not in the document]
caspervonb: Its "filter" in the go client but, dubbed it preprocessors.
There's one on each of the channel.

src/jetstream_push_subscription.rs
caspervonb: Want to change `process` calls to error out @Jarema?

src/client.rs
caspervonb: Renamed to _preprocess_ on both ends (before send in client and internal recv filter in subscription).

src/jetstream_push_subscription.rs
caspervonb: Changed to error out if ack fails.

Jarema:
When Derek comments are resolved: LGTM.

Error types changes should be handled in separate PR.

Jarema (approved):

src/jetstream_types.rs
caspervonb: Wrote and added serde_nanos to handle serialisation, works for both chrono and std durations but went with std.

derekcollison:
Are you both ok with the changes here? I will not have the time to do a detailed walkthrough. If you both are good to go then that is good for me.

Jarema:
> Are you both ok with the changes here? I will not have the time to do a detailed walkthrough. If you both are good to go then that is good for me.

We went through few iterations with this PR and for me it's good to go.

PR #253, opened 2021-11-11
Some additions..

derekcollison:
Added in async error callback with defaults eprintln!
Made Message is_no_responders() public for normal use.

Signed-off-by: Derek Collison <derek@nats.io>

### DIFF
--- a/src/options.rs
+++ b/src/options.rs
@@ -15,12 +15,14 @@ use std::cmp;
 use std::convert::TryInto;
 use std::fmt;
 use std::io;
+use std::io::Error;
 use std::path::{Path, PathBuf};
 use std::sync::Arc;
 use std::time::Duration;
 
 use crate::auth_utils;
 use crate::secure_wipe::SecureString;
+use crate::Client;
 use crate::Connection;
 
 /// Connect options.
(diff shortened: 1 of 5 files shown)


src/options.rs
```diff
@@ -667,3 +693,26 @@ impl ReconnectDelayCallback {
         self.0(reconnects)
     }
 }
+
+pub(crate) struct ErrCallback(Option<Box<dyn Fn(Error) + Send + Sync + 'static>>);
```
caspervonb, line 697: Nit (`Err` and `std::Error` are different things)

```suggestion
pub(crate) struct ErrorCallback(Option<Box<dyn Fn(Error) + Send + Sync + 'static>>);
```

src/options.rs
```diff
@@ -667,3 +693,26 @@ impl ReconnectDelayCallback {
...
+impl ErrCallback {
+    pub fn call(&self, client: &Client, err: Error) {
+        if let Some(callback) = self.0.as_ref() {
+            callback(err);
+        } else {
+            let si = client.server_info();
+            eprintln!("{} on connection [{}]", err, si.client_id);
```
caspervonb, line 704: Maybe we should call into `log`(which is configurable) instead of printing directly to stderr here.

caspervonb (commented):
Minor nits

src/options.rs
derekcollison: This is a default version, we could call into log though if we already have it in the client..

derekcollison: So with log.error! default does not print anything and I want the default handler, meaning if you do not set one, to print to stderr when bad things happen.

caspervonb (approved):

PR #259, opened 2021-11-16
Use temporary directories for storage

caspervonb:
Quotes were resulting in a directory named `./$SD` directory to be created and used on Unix for storage, which leads to racing between the tests and frequent failures.

This makes use of the `--store_dir` cli option instead.

### DIFF
--- a/tests/util/mod.rs
+++ b/tests/util/mod.rs
@@ -87,12 +87,13 @@ impl Server {
 pub fn run_server(cfg: &str) -> Server {
     let id = nuid::next();
     let logfile = env::temp_dir().join(format!("nats-server-{}.log", id));
-    let sd = env::temp_dir().join(format!("store-dir-{}", id));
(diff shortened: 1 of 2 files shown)


tests/util/mod.rs
```diff
@@ -87,12 +87,13 @@ impl Server {
...
     let id = nuid::next();
     let logfile = env::temp_dir().join(format!("nats-server-{}.log", id));
-    let sd = env::temp_dir().join(format!("store-dir-{}", id));
+    let store_dir = env::temp_dir().join(format!("store-dir-{}", id));
 
     // Always use dynamic ports so tests can run in parallel.
     // Create env for a storage directory for jetstream.
     let mut cmd = Command::new("nats-server");
-    cmd.env("SD", sd)
+    cmd.arg("--store_dir")
```
derekcollison, line 95: can be just -sd too.

derekcollison (approved):
LGTM

PR #263, opened 2021-11-19
Workflow improvements

caspervonb:
- Remove jetstream job
- Don't use rustup to re-install clippy
- Normalises jobs and step naming splitting them into two groups: tests and checks.


caspervonb:
Maybe make use ofactions-rs/toolchain instead of `rustup` while we're at it?

Jarema (approved):

PR #271, opened 2021-12-19
fix: publicly export handler struct

JakeStanger:
Currently the `Handler` struct returned from `Subscription#with_handler` is public but not exported as part of the library. This exports it so the handler can be used as a return type etc.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -247,7 +247,7 @@ use regex::Regex;
 pub use jetstream::JetStreamOptions;
 pub use message::Message;
 pub use options::Options;
-pub use subscription::Subscription;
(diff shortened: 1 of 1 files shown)


Jarema (commented):
Thanks for the PR @JakeStanger!

Please fix `cargo fmt` (it expects `{Handler, Subscription}` ) and on the occasion of that amend, commit changing message to imperative mood for consistency (i.e. `Publicly export Handler`).

Otherwise, LGTM!

JakeStanger:
Oops - sorry for missing that `fmt` off earlier. Changes pushed :)

Jarema (approved):
All good.

Thanks for your contribution!

PR #307, opened 2022-01-20
Change jetstream log level to debug

### LINKED ISSUE #306: Log statements can be confusing for errors that are redirected
This came up in Slack.

For example, key-value's `get` sends a "get message" request.
This then logs that the message doesn't exist but this is expected for `get` and we catch this and map it to None.

caspervonb:
Closes #306


Jarema (approved):
LGTM

PR #334, opened 2022-02-23
Fix linter warnings

caspervonb:


nats-experimental/src/lib.rs
```diff
@@ -92,7 +92,7 @@ impl ServerInfo {
 #[derive(Clone, Debug)]
 pub enum ServerOp {
     Ok,
-    Info(ServerInfo),
+    Info(Box<ServerInfo>),
```
caspervonb, line 95: I boxed this to avoid padding the entire enum.

Jarema (approved):
LGTM

PR #337, opened 2022-03-03
Bump rustls to `0.20.4`

hyche:
This allows building `nats` on Apple M1 devices, because `rustlts` v0.19 depends on `ring` v0.16.12, which build failed on M1.
Issue: https://github.com/briansmith/ring/issues/1163


caspervonb:
Both me and @Jarema develop on M1 Macs so not sure if this is a current problem?

hyche:
I figured out this can be resolved with `cargo update`, this command can update deps to the latest minor version. If you start a project from scratch then this is not a problem, but there are some old projects that contains a `Cargo.lock` file which locks in old version of `ring`, then this is a problem.

PR #352, opened 2022-04-08
Consume server info on connect

caspervonb:
Consumes the first op on connect to prevent silent failures


Jarema (commented):
We're reading first op while connecting anyway. Any reason to have it once more here?

caspervonb:
We are now, we weren't 🙃

PR #374, opened 2022-04-19
Send unsubscribe op on subscription receiver closed

caspervonb:
Alternative to #363


caspervonb:
Closing in favour of https://github.com/nats-io/nats.rs/pull/363, which turns out to be leaner.

PR #381, opened 2022-04-20
Improve workflow caching

caspervonb:


caspervonb:
Seems we are hitting cache, I'll leave this at just bumping the action version for now.

Jarema (approved):
LGTM

PR #392, opened 2022-04-27
Handle connections in a `Connector` type

caspervonb:


Jarema (changes requested):
General idea is fine, but see tests.

caspervonb:
Was dependent on https://github.com/nats-io/nats.rs/pull/390

Jarema (approved):
LGTM

PR #403, opened 2022-04-30
Discover additional servers via INFO

caspervonb:


async-nats/src/lib.rs
```diff
@@ -493,7 +493,14 @@ impl Connector {
 
         for server_addr in &self.server_addrs {
             match self.try_connect_to(server_addr).await {
-                Ok(inner) => return Ok(inner),
+                Ok((info, connection)) => {
+                    for url in &info.connect_urls {
+                        self.server_addrs.push(url.parse()?);
```
Jarema, line 498: We should check for duplicates, as it will duplicate used address and further duplicate everything on each reconnect.

Jarema (changes requested):

async-nats/src/lib.rs
caspervonb: sgtm, want to just check for contains or make it a set?

Jarema: Inserting to vec with checking for duplicates via contains sounds like hashSet with extra steps :).

Jarema: If I recall correctly, the sync client coutns the number of reconnects for each URL.
Would be good to checkout Go Client if it does the same. If yes- it should be a HashMap with key being URL and value - number of attempts.

async-nats/src/lib.rs
```diff
@@ -501,9 +509,33 @@ impl Connector {
...
+                cmp::min(Duration::from_millis(2_u64.saturating_pow(exp)), max)
+            };
+
+            *server_attempts += 1;
+            sleep(duration).await;
+
+            match self.try_connect_to(&server_addr).await {
+                Ok((info, connection)) => {
+                    for url in &info.connect_urls {
+                        self.attempts.insert(url.parse()?, 0);
```
Jarema, line 530: This will reset attempts for any erroneous server that is in the info list (and those URLs are prone to errors as the may be internal and not accessible from the Client)

async-nats/src/lib.rs
```diff
@@ -501,9 +509,33 @@ impl Connector {
...
+                    }
+
+                    let server_attempts = self.attempts.get_mut(&server_addr).unwrap();
+                    *server_attempts = 0;
```
Jarema, line 534: isn't line 524 `server_attempts` still alive here?

Jarema (changes requested):

async-nats/src/lib.rs
caspervonb: It can't be, would be double borrow for self.

async-nats/src/lib.rs
```diff
@@ -473,11 +473,25 @@ impl Connection {
 
 /// Maintains a list of servers and establishes connections.
 pub(crate) struct Connector {
-    server_addrs: Vec<ServerAddr>,
+    /// A map of servers and number of connect attempts.
+    attempts: HashMap<ServerAddr, usize>,
```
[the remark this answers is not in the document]
caspervonb: Name yanked from the nats crate.

async-nats/src/lib.rs
caspervonb: Semantics yanked from nats/connector.rs

Jarema: Not really. You got all advertised URLs here, not only the one that actually works. Is fine to reset only the one we connected to.

async-nats/src/lib.rs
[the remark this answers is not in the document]
Jarema: Let me paraphrase @caspervonb : let's not justify things with "it was like this before" 😆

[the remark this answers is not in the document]
caspervonb: Had a feeling you would quote me here 😁

Jarema:
Let me know when agreed changes are in.

Jarema (approved):
LGTM!

PR #407, opened 2022-05-02
Sleep between connection attempts

caspervonb:


Jarema (changes requested):
Let’s make this a proper backoff.
Customizing and disabling it can be followed up with another PR though.

Jarema (approved):
LGTM

PR #418, opened 2022-05-04
Fix flaky connection callback test

caspervonb:
Edit: this wasn't it, the server just doesn't get killed *sometimes*.


caspervonb:
This wasn't it, the server just doesn't get killed sometimes.

PR #421, opened 2022-05-06
Fix/unmute

stevelr:
When porting some nats client code to async-nats, I found the greatest amount of change required was due to the need for the Client to be `mut`. But it turns out that most of the time it doesn't. This change removes unneeded `mut` in function signatures.


stevelr:
(sorry I probably should have rebased this from nats-io/nats.rs main but it's a branch from the jwt auth branch) This is the commit that removed the mut:  https://github.com/nats-io/nats.rs/pull/421/commits/5c65e10c08b9e5a45ea33508e4a8e5279226cce8

caspervonb:
lgtm

PR #424, opened 2022-05-07
Handle `TryFlush` as a command

caspervonb:
This inlines Command::TryFlush in the command handler, rather than doing it during operation encoding.


Jarema (approved):
LGTM

PR #425, opened 2022-05-07
Fix all clippy warnings

caspervonb:
Fixed all linter warnings reported with:

```
cargo clippy --examples --tests --all-features -- --deny clippy::all 
```


Jarema (approved):
LGTM

PR #430, opened 2022-05-09
Prefer re-exports from futures crate

caspervonb:
Drops us having two manage the two extra dependencies explicitly, they're re-exported from the futures crate.


Jarema (approved):
LGTM (assuming green CI)

PR #445, opened 2022-05-17
Add test case for connecting to domains

caspervonb:


caspervonb:
Was expecting this to fail based on what we're seeing in https://github.com/nats-io/nats.rs/pull/428.

Jarema (approved):
LGTM

PR #483, opened 2022-06-10
Use a concurrency group to prevent action backlog

caspervonb:
Adds a concurrency group to prevent force pushes and similar events on the same pull request to fill up the backlog for hours (benchmark takes quite some time once it starts filling up).


Jarema (approved):
LGTM

PR #487, opened 2022-06-11
Make `jetstream::Context::create_stream` return `Stream`

caspervonb:
The two paths of get_or_create_stream returns different types, which is fairly unexpected.

This changes `jetstream::Context::create_stream` to return a `Stream`. If we need a non-binding variant we can add `insert_stream` which returns `stream::Info` rather than `Stream`.


Jarema (approved):
LGTM

PR #488, opened 2022-06-11
Rename `BatchIter` to `Batch`

caspervonb:
Prior art in both std and our old nats implementation tends to lean towards nouns without a suffix.


Jarema (approved):
LGTM

PR #508, opened 2022-06-16
Rename `Process` stream to `Sequence`

caspervonb:
Alternative names include but are not limited to series, serial, linear, etc.


Jarema (approved):
LGTM

PR #516, opened 2022-06-17
Normalize the examples heading

caspervonb:
For consistency without trailing colon ala std lib as it is rendered with a heading tag in rustdoc.


Jarema (approved):
LGTM

PR #524, opened 2022-06-21
Normalize error type used in subscribe methods

caspervonb:


Jarema (approved):
LGTM assuming green CI.

PR #530, opened 2022-06-23
Seperate benchmark workflow from test workflow

caspervonb:
Benchmarks is taking too many cycles on the CI.
This splits them into their own workflow that only runs against main.


Jarema (approved):
lgtm,

needs follow up with running only against release though.

PR #533, opened 2022-06-25
Inline subscription calls

caspervonb:


caspervonb:
> Hm, i think its better idea to fins a good internal rather than inline those calls?

Not much duplication here, since they're both just sending a command to a channel.

Jarema (approved):
lgtm

PR #535, opened 2022-06-25
Testing pull stream rework

caspervonb:
Trying to figure out the reason for the failures we're seeing on CI with macOS in #529


caspervonb:
Discovery: sending requests after timeout messages ends up with server not sending us any more messages.

PR #556, opened 2022-07-08
Use correct flush_timeout operation in doc test

krady21:
Hi. I've been going through the code as I've been looking to contribute and I noticed the wrong function being used in one of the doc tests.


caspervonb (approved):
Oops, nice catch! thank you 💯

PR #566, opened 2022-07-12
Rename push consumer `Stream` iterator to `Messages`

caspervonb:


Jarema (approved):
Yeah, let's have it in  for consistency.

LGTM.

PR #567, opened 2022-07-12
Remove fault injection run from workflow

caspervonb:
We don't make much use of this in async, takes up runner time.


Jarema (approved):
lgtm

PR #598, opened 2022-08-07
Inline header module from http crate

caspervonb:


caspervonb:
Closing in favour of #629

PR #666, opened 2022-10-18
Bump rustls-pemfile version

paulgb:
`async_nats` already depends on `rustls-native-certs v0.6.2`, which depends on `rustls-pemfile 1.0.1`. This change bumps the version of `rustls-pemfile` depended on directly on that version, so that the direct dependency matches the transitive dependency, preventing the compiler from compiling two different versions of that crate.


Jarema (approved):
Yes, that's a good idea to bump that.
Thanks for also updating dependencies.md 🙂 

LGTM!

PR #680, opened 2022-10-31
Use subscriber term consistently

caspervonb:
Minor documentation change to use the term subscriber over sub consistently.


Jarema (approved):
LGTM

PR #681, opened 2022-10-31
Merge periodic ping into connection handler loop

caspervonb:
Connection handler is supposed to exit when all senders are dropped, ping and flush tasks currently keep it alive.


Jarema (approved):
LGTM

PR #687, opened 2022-11-03
Merge periodic flush into connection handler loop

caspervonb:
Command sender is no longer cloned, so the loop should now break when all senders (client, subscriber, etc) are closed.

Depends on #681


Jarema (changes requested):
code is fine, but please add tests for dropping the connection.

caspervonb:
Fun thing to test, this is all internal 😁

Jarema (approved):
LGTM

PR #688, opened 2022-11-03
Update K/V purge documentation verbiage

corbinu:
Clarifies K/V purge documentation to record that it is a removal of entries not the key itself from the bucket.

### DIFF
--- a/async-nats/src/jetstream/kv/mod.rs
+++ b/async-nats/src/jetstream/kv/mod.rs
@@ -543,7 +543,7 @@ impl Store {
         Ok(())
     }
 
-    /// Purges the given key, destructively removing everything from the bucket.
+    /// Purges the given key history, destructively removing all key entries from the bucket.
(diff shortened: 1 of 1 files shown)


Jarema (approved):
LGTM!

async-nats/src/jetstream/kv/mod.rs
```diff
@@ -543,7 +543,7 @@ impl Store {
         Ok(())
     }
 
-    /// Purges the given key, destructively removing everything from the bucket.
+    /// Purges the given key history, destructively removing all key entries from the bucket.
```
caspervonb, line 546: What about something ala

```suggestion
    /// Purges all the revisions of a key destructively, leaving a single purge marker in-place.
```

corbinu: Sure will update

Issue #693, opened 2022-11-04
Can't run clippy on CI from outside fork


caspervonb:
CI started failing with access permission errors

```
Clippy results: 0 ICE, 0 errors, 0 warnings, 0 notes, 0 help
[837](https://github.com/nats-io/nats.rs/actions/runs/3394399147/jobs/5642915730#step:5:839)
Error: Unable to create clippy annotations! Reason: HttpError: Resource not accessible by integration
[838](https://github.com/nats-io/nats.rs/actions/runs/3394399147/jobs/5642915730#step:5:840)
Warning: It seems that this Action is executed from the forked repository.
[839](https://github.com/nats-io/nats.rs/actions/runs/3394399147/jobs/5642915730#step:5:841)
Warning: GitHub Actions are not allowed to create Check annotations, when executed for a forked repos. See https://github.com/actions-rs/clippy-check/issues/2 for details.
[840](https://github.com/nats-io/nats.rs/actions/runs/3394399147/jobs/5642915730#step:5:842)
Posting clippy checks here instead.
[841](https://github.com/nats-io/nats.rs/actions/runs/3394399147/jobs/5642915730#step:5:843)
Error: Clippy had exited with the 101 exit code
```

caspervonb:
Adding the GitHub key to the workflow should fix the annotations.
e.g

```yaml
jobs:
  ...:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v1
      - run: rustup component add clippy
      - uses: actions-rs/clippy-check@v1
        with:
          token: ${{ secrets.GITHUB_TOKEN }}
          args: ...
```

corbinu:
Sure!

corbinu:
@caspervonb It looks like this is an unsolved issue with clippy-check. The solution seems to be to use scherermichael-oss/action-has-permission to check if the token has write permissions and fallback to running clippy in the fork without annotations.

corbinu:
https://github.com/rust-lang/rust-clippy/issues/9781

PR #694, opened 2022-11-04
CI only write annotations from clippy when write permissions exist

corbinu:
Update linting in CI to only use clippy-check (and therefore try to write annotations) when the write permission is present on the github token.

This solves https://github.com/nats-io/nats.rs/issues/693 due to clippy-check not being able to run in forks of a project https://github.com/actions-rs/clippy-check/issues/2

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -123,11 +123,24 @@ jobs:
           components: clippy
           override: true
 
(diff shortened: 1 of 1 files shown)


caspervonb:
Need to get CI passing again, going to close this in favor of https://github.com/nats-io/nats.rs/pull/695.

PR #737, opened 2022-11-21
Reset flush interval after ping forces flush

caspervonb:
Resets the flush interval after ping forces a flush to avoid redundant flush calls


Jarema (approved):
LGTM

PR #742, opened 2022-11-22
Fix-up message header documentation

caspervonb:
Fixup a minor documentation mention of the http crate, as we no longer use the http crate for headers.


Jarema (approved):
LGTM

PR #747, opened 2022-11-29
Always reset periodic flush interval after manual flush

caspervonb:


async-nats/src/lib.rs
```diff
@@ -298,46 +298,48 @@ pub(crate) struct ConnectionHandler {
...
         loop {
             select! {
-                _ = ping_interval.tick().fuse() => {
+                _ = self.ping_interval.tick().fuse() => {
                     self.pending_pings += 1;
                     if let Err(_err) = self.connection.write_op(ClientOp::Ping).await {
                         self.handle_disconnect().await?;
                     }
 
-                    self.connection.flush().await?;
```
caspervonb, line 336: Forgot to stick flush back in here

Jarema (approved):
LGTM

PR #776, opened 2022-12-19
Use stable channel for clippy

caspervonb:
New version shipped over the last couple of days, clippy should be working again on stable


Jarema (approved):
LGTM

PR #824, opened 2023-02-02
fix(consumers): Fixes duplicate consumer creation

thomastaylor312:
The spawned task watching the consumer and recreating was never aborted when the consumer was dropped


Jarema (approved):
LGTM

PR #844, opened 2023-02-23
Make `Context::request` return `IntoFuture` builder

n1ghtmare:
Adds timeout to request and makes it a builder, also completes https://github.com/nats-io/nats.rs/pull/782


Jarema (approved):
LGTM

PR #870, opened 2023-03-09
Updated ObjectInfo struct definition

kristianhasselknippe:
When using the jetstream object store, we I was not able to read files as the struct definition for ObjectInfo didn't match the data.

The name of the modified field was incorrect, and judging from the go implementation, the `digest` and `deleted` fields seems to be optional.


caspervonb:
Thank you very much @kristianhasselknippe.
Added a test and merged in https://github.com/nats-io/nats.rs/pull/895, credited you as the author.

PR #873, opened 2023-03-10
Fix linting errors and suggestions

n1ghtmare:
Fixes linting errors related to the `derive_default_enum` feature, more info [here](https://rust-lang.github.io/rfcs/3107-derive-default-enum.html). 
Also, incorporated some other linter suggestions (getting rid of an unnecessary `into_iter` call).


Jarema (approved):
LGTM!

PR #878, opened 2023-03-14
Prefer returning `&str` over String where applicable

paolobarbolini:
Split out of #876 because of breaking changes to the public API


Jarema (approved):
LGTM!

PR #894, opened 2023-03-23
Remove collect to Vec in try_read_op

YaZasnyal:
Here are changes that were suggested in https://github.com/nats-io/nats.rs/issues/889#issue-1634670217.


Jarema (approved):
LGTM!

PR #915, opened 2023-04-07
Bump `tokio-rustls` version and add `webpki` as a dependency

n1ghtmare:
Addresses: #882 

`tokio-rustls 0.24` uses `rustls 0.21.0`

Note that `webpki` had to be updated as well since it has been removed from the public API of the new version of `tokio-rustls` (For reference: https://github.com/tokio-rs/tls/pull/82)

### DIFF
--- a/async-nats/Cargo.toml
+++ b/async-nats/Cargo.toml
@@ -27,7 +27,7 @@ http = "0.2.9"
 tokio = { version = "1.25.0", features = ["macros", "rt", "fs", "net", "sync", "time", "io-util"] }
 itoa = "1"
 url = "2"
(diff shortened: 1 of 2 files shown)


async-nats/Cargo.toml
```diff
@@ -39,6 +39,7 @@ base64 = "0.21"
 tokio-retry = "0.3"
 ring = "0.16"
 rand = "0.8"
+webpki = "0.22.0"
```
paolobarbolini, line 42: Wrong version. rustls is using it's own fork

```suggestion
webpki = { package = "rustls-webpki", version = "0.100.0", features = ["alloc", "std"] }
```

n1ghtmare: Thank you!

Jarema (approved):
LGTM!

PR #922, opened 2023-04-13
Impl From<HeaderName> for String

matthiasbeyer:
Same as #921 - maybe there's a good reason why this isn't implemented. If so, it is not documented.

:smile:

### DIFF
--- a/async-nats/src/header.rs
+++ b/async-nats/src/header.rs
@@ -311,6 +311,12 @@ impl FromStr for HeaderName {
     }
 }
 
+impl From<HeaderName> for String {
+    fn from(hn: HeaderName) -> Self {
+        hn.value
+    }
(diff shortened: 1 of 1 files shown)


matthiasbeyer:
In this case it might still be an idea to provide some `to_string()` implementation for users, if they want to go the extra step and allocate a `String`, won't it?

matthiasbeyer:
I don't know. Is this something that can be shown to a user of an application (or a developer that does some network debugging or whatever)?

And can `ByteString` (not sure whether it would come from) implement `Display` as well?

Maybe `AsRef<str>` would be enough...

matthiasbeyer:
Closing as superseded by #924 

@caspervonb :+1: thanks!

PR #924, opened 2023-04-16
Implement `fmt::Display` for `HeaderName`

caspervonb:
Implements `fmt::Display` for `HeaderName`, which in turn provides a blanket implementation for `ToString`.


Jarema (approved):
LGTM

Issue #923, opened 2023-04-13
Constant CPU usage with axum/tower & ConnectOptions


luizfonseca:
Make sure that these boxes are checked before submitting your issue -- thank you!

 - [x] Included below version and environment information
 - [x] Included a [Minimal, Complete, and Verifiable example] (https://stackoverflow.com/help/mcve)

#### NATS version (`grep 'name = "nats"' Cargo.lock -A 1`)
0.29.0

#### NATS server version
v2.9.15

#### rustc version (`rustc --version` - we support Rust 1.41 and up)
1.68.2

#### OS/Container environment:
MacOS Ventura 13.1, M1 mac

#### Steps or code to reproduce the issue:

Hello there. I have been learning Rust and in the process of learning NATS as well. 
I have a mock deployment that I manage locally mimicking a production environment using `dokku`
and noticed something particularly odd about CPU usage.

Important: code runs with `--release`.

1. Have an `axum` and `tower` server with a shared state using `Arc`:
```rust
use std::{net::SocketAddr, sync::Arc};
use axum::{routing::get, Router};
use async_nats::ConnectOptions;

struct SharedState {
   nats: async_nats::Client,
}

#[tokio::main]
async fn main () -> Result<(), anyhow::Error> {
   let nats_client = ConnectOptions::with_user_and_password("user", "pw").connect("url").await?;
   
   // Creates shared state for all handlers
   let shared_state = Arc::new(SharedState {
      nats: nats_client
   });
   
   let app = Router::new()
                   .route("/", get { || async { "Test" } })
                   .with_state(shared_state);
   
    let addr = SocketAddr::from(([0, 0, 0, 0], 3030));
     axum::Server::bind(&addr)
        .serve(app.into_make_service_with_connect_info::<SocketAddr>())
        .await?;
}

```

2. Observe as CPU usage hovers within 2-3% during the server execution, even without subscribers/publishers

![img](https://user-images.githubusercontent.com/249782/231839531-e59f38b1-8d2a-4ab2-b026-d1e2460c221b.png)

_increase in CPU usage after installing `async-nats`. Blue in the graph showcases the issue._

#### Expected result:
- CPU behaves normally when idle (see screenshot above for periods without the issue)

#### Actual result:

- While on idle (no messages or subscribers), CPU sits at 2-3% all the time. 
- Occurs when using the `async-nats` crate 
- When reverting to `nats 0.24.0`, the issue goes away and CPU usage is back to < 1%.

[...]

PR #932, opened 2023-04-24
Reset ping interval on incoming and outgoing messages

### LINKED ISSUE #931: Reduce outgoing pings
Go client is quite stringent with its pings, we could be too.
Whenever we receive traffic, we could set an instant representing the last inbound message.
During the ping interval, we only send the ping if enough time has passed since last inbound message.

caspervonb:
Closes #931 
Partially resolves https://github.com/nats-io/nats.rs/issues/923


Jarema (approved):
LGTM

PR #933, opened 2023-04-25
Reduce redundant flush wakes

### LINKED ISSUE #923: Constant CPU usage with axum/tower & ConnectOptions

caspervonb:
Closes #923 
See https://github.com/nats-io/nats.rs/issues/905

### DIFF
--- a/async-nats/src/lib.rs
+++ b/async-nats/src/lib.rs
@@ -304,7 +304,8 @@ pub(crate) struct ConnectionHandler {
     max_pings: usize,
     info_sender: tokio::sync::watch::Sender<ServerInfo>,
     ping_interval: Interval,
-    flush_interval: Interval,
+    flush_period: Duration,
+    flush_wanted: bool,
 }
 
 impl ConnectionHandler {
@@ -318,9 +319,6 @@ impl ConnectionHandler {
         let mut ping_interval = interval(ping_period);
         ping_interval.set_missed_tick_behavior(MissedTickBehavior::Delay);
 
-        let mut flush_interval = interval(flush_period);
-        flush_interval.set_missed_tick_behavior(MissedTickBehavior::Skip);
-
         ConnectionHandler {
             connection,
             connector,
@@ -329,7 +327,8 @@ impl ConnectionHandler {
             max_pings: 2,
             info_sender,
             ping_interval,
-            flush_interval,
+            flush_period,
+            flush_wanted: false,
         }
     }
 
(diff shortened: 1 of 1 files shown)


caspervonb:
Brings idle usage down from a 8-6% to 0% when idle (e.g just waiting for a socket read from subscription).

## Before

image

## After

image

async-nats/src/lib.rs
```diff
@@ -338,6 +337,12 @@ impl ConnectionHandler {
         mut receiver: mpsc::Receiver<Command>,
     ) -> Result<(), io::Error> {
         loop {
+            let flush_sleep = if self.flush_wanted {
+                tokio::time::sleep(self.flush_period).boxed()
```
paolobarbolini, line 341: This will create a lot of allocator noise. What if we made the `Interval` very high (a few days long) when we don't want any flushes, and lowered it back to the configured value when we want it back?

paolobarbolini: Or we could make our own `Interval` wrapper around [`Sleep`](https://docs.rs/tokio/latest/tokio/time/struct.Sleep.html#method.reset)

> This function can be called both before and after the future has completed.

caspervonb: > This will create a lot of allocator noise.
Aye, but we can take it out again with a little bit of work.

> Or we could make our own Interval wrapper around [Sleep](https://docs.rs/tokio/latest/tokio/time/struct.Sleep.html#method.reset)

Yeah short rundown is that we'd want to return pending when flush not wanted, store the waker, wake up when flush wanted is.

caspervonb:
Closing in favor of #1070

PR #934, opened 2023-04-25
Replace flush interval with a fused sleep

caspervonb:
Alternative to #933


caspervonb:
Marginally better than #933, still can't reach in and reset the sleep however.

caspervonb:
Closing in favor of #1070

PR #935, opened 2023-04-25
Reduce usage of the futures crate

paolobarbolini:
Removes usage of `futures` macros. Considering this crate is very much married to the `tokio` ecosystem, it didn't make sense to also _sometimes_ use the `futures` version of things. `tokio::select` was already being used somewhere else


caspervonb:
I'm inclined to go the other direction, prefer futures crate over tokio, prefer std over futures.
Will discuss with Thomasz offline.

dodomorandi:
Would you consider introducing the usage of [`futures-concurrency`](https://docs.rs/futures-concurrency/latest/futures_concurrency/)? Surely changing the abstraction is not a trivial task (especially for `select!` -> `merge`), but I think it could be worth considering the idea.

caspervonb:
> Would you consider introducing the usage of [futures-concurrency](https://docs.rs/futures-concurrency/latest/futures_concurrency/)? Surely changing the abstraction is not a trivial task (especially for select! -> merge), but I think it could be worth considering the idea.

Never outright against something, will look into it.

PR #952, opened 2023-05-01
Add subject type

### LINKED ISSUE #950: Use `Bytes` with a string interface for subjects to avoid allocations.
### Use case

In cases where you know what the subject is at compile-time, we incur a lot of redundant overhead from String allocations, e.g:

```rust
for i in 0..1_000_000 {
  context.publish("some_known_subject".into(), format!("{}", i).into()).await?;
}
```

### Proposed change

Introduce a `Subject` type which is constructible from a `&'static str` reference.

This type should safely be able to implement a `Deref` to str in order to maintain a string-like interface. This can be enforced by only allowing utf8 strings to become subjects.

```rust
#[derive(Debug, Clone, Eq, PartialEq, Ord, PartialOrd, Hash)]
pub struct Subject {
    inner: Bytes,
}

impl Subject {
    #[inline]
    pub const fn from_static(s: &'static str) -> Subject {
        Subject {
            inner: Bytes::from_static(s.as_bytes()),
        }
    }

    pub fn from_utf8(vec: Vec<u8>) -> Result<Subject, FromUtf8Error> {
      // ...
    }
}

impl ops::Deref for Subject {
    type Target = str;

    #[inline]
    fn deref(&self) -> &str {
        let b: &[u8] = self.inner.as_ref();
        unsafe { str::from_utf8_unchecked(b) }
    }
}
```

### Who benefits from the change(s)?

Users wanting to minimize subject allocations

### Alternative Approaches

_No response_

caspervonb:
Adds a bytes based immutable string type for subjects.
Goal is to have the same interface as String, minus the mutability.

- [x] Make it compile
- [x] Make tests compile
- [x] Make examples compile
- [x] Update documentation and make it compile
- [ ] Take out redundant allocations
- [ ] Forward `Bytes` directly where applicable

Closes #950


Jarema:
We should validate subjects when publishing differently than when subscribing.

When subscribing, we only validate for a valid subject.
When publishing, we can't use wildcards.

Not sure when the check should happen yet.

caspervonb:
Seeing a slight performance bump from main

```
async-nats: publish messages amount/32                                                                           
                        time:   [52.064 µs 52.322 µs 52.550 µs]
                        thrpt:  [1.9029 Melem/s 1.9113 Melem/s 1.9207 Melem/s]
                 change:
                        time:   [-3.7262% -0.2156% +2.1908%] (p = 0.91 > 0.05)
                        thrpt:  [-2.1438% +0.2161% +3.8704%]
                        No change in performance detected.
Found 8 outliers among 30 measurements (26.67%)
  1 (3.33%) low severe
  4 (13.33%) low mild
  2 (6.67%) high mild
  1 (3.33%) high severe
async-nats: publish messages amount/1024                                                                           
                        time:   [101.89 µs 102.36 µs 103.02 µs]
                        thrpt:  [970.68 Kelem/s 976.95 Kelem/s 981.45 Kelem/s]
                 change:
                        time:   [-0.5065% +0.0175% +0.5152%] (p = 0.95 > 0.05)
                        thrpt:  [-0.5126% -0.0175% +0.5090%]
                        No change in performance detected.
async-nats: publish messages amount/8192                                                                            
                        time:   [684.27 µs 690.45 µs 699.05 µs]
                        thrpt:  [143.05 Kelem/s 144.83 Kelem/s 146.14 Kelem/s]
                 change:
                        time:   [-8.6033% -5.2521% -2.8634%] (p = 0.00 < 0.05)
                        thrpt:  [+2.9478% +5.5432% +9.4131%]
                        Performance has improved.
 ```

**Comment by caspervonb** [comment]
Instrumented allocations memory (left is this branch, right is main)

image

File: async-nats/src/lib.rs [hunk 1]
```diff
@@ -445,7 +447,7 @@ impl ConnectionHandler {
                 length,
             } => {
                 if let Some(subscription) = self.subscriptions.get_mut(&sid) {
-                    let message = Message {
+                    let message: Message = Message {
```
n1ghtmare, line 450:
```suggestion
                    let message = Message {
```
This change might not be needed (most probably added by me, during rebase...).

n1ghtmare (approved):
LGTM! Tiniest of tiny nits...

MattesWhite:
> We should validate subjects when publishing differently than when subscribing.
> 
> When subscribing, we only validate for a valid subject. When publishing, we can't use wildcards.
> 
> Not sure when the check should happen yet.

Just wanted to through in a reminder that #596 exists 😄. So the only thing I really would like to see when a `Subject` type is introduced is the possibility to iterate over tokens of a subject. This makes it way more comfortable to check subjects of a messages from a wildcard subscription for what was actually sent, e.g.:

- Subscription to `my-app.*.data`
- Receive message from `my-app.1234.data`
- Use `Subject::tokens().nth(1)` to be able to parse the app's ID

All in all, a nice implementation 👍 I really like the use of `Bytes` to prevent allocations.

Jarema (approved):
LGTM!

PR #953, opened 2023-05-03
Make `Command` private

caspervonb:
This type is just used internally to communicate with the internal connection handler and not publicly useful to anyone.


Jarema (approved):
LGTM

PR #956, opened 2023-05-04
Disconnect if pending pings is more than max pings

### LINKED ISSUE #955: Unresponsive server with pings do not trigger disconnect
### NATS version

0.29.0

### rustc version

rustc 1.71.0-nightly (c609da59d 2023-04-18)

### OS/Container environment

MacOS

### Steps or code to reproduce the issue

Regression introduced when moving periodic pings to intervals.
We no longer trigger disconnect when there are more than 2 pongs in-flight.

### Expected result

Connection should disconnect.

### Actual result

Connection keeps accumulating pending pings, despite possibly being stale.

caspervonb:
Closes #955

### DIFF
--- a/async-nats/src/lib.rs
+++ b/async-nats/src/lib.rs
@@ -340,6 +340,15 @@ impl ConnectionHandler {
             select! {
                 _ = self.ping_interval.tick().fuse() => {
                     self.pending_pings += 1;
+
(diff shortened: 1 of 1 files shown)


Jarema (approved):
LGTM

PR #967, opened 2023-05-12
Add static representation for custom header names

caspervonb:
Adds static representation for custom header names.

Blocked-on #946


Jarema (approved):
LGTM!

PR #975, opened 2023-05-25
Derive `Clone` for `Message`

mgrachev:
To have ability to send a message via channels like [tokio::sync::broadcast](https://docs.rs/tokio/latest/tokio/sync/broadcast/fn.channel.html).


caspervonb (approved):
lgtm

PR #989, opened 2023-06-05
Improve `connection` module documentation

n1ghtmare:


async-nats/src/connection.rs
```diff
@@ -56,6 +59,7 @@ pub(crate) struct Connection {
 /// Internal representation of the connection.
 /// Holds connection with NATS Server and communicates with `Client` via channels.
 impl Connection {
+    /// Attempt to read a server operation from the connection buffer.
```
caspervonb, line 62:
```suggestion
    /// Attempts to read a server operation from the read buffer.
    /// Returns `None` if there is not enough data to parse an entire operation.
```

async-nats/src/connection.rs
```diff
@@ -316,6 +320,7 @@ impl Connection {
     }
 
     // TODO: do we want an custom error here?
+    /// Read a server operation from the connection stream asynchronously.
```
caspervonb, line 323:
```suggestion
    /// Read a server operation from read buffer.
    /// Blocks until an operation can be parsed.
```

async-nats/src/connection.rs
```diff
@@ -332,6 +337,7 @@ impl Connection {
         }
     }
 
+    /// Write a client operation to the connection stream asynchronously.
```
caspervonb, line 340:
```suggestion
    /// Writes a client operation to the write buffer.
```

async-nats/src/connection.rs
```diff
@@ -438,6 +444,7 @@ impl Connection {
         Ok(())
     }
 
+    /// Flush the connection buffer, ensuring that all pending data is written to the server.
```
caspervonb, line 447:
```suggestion
    /// Flush the  write buffer, sending all pending data down the current write stream.
```

Jarema (approved):
LGTM

PR #994, opened 2023-06-08
Add inactive threshold for pull consumer

paulgb:
This exposes the `InactiveThreshold` parameter of a push stream, similar to how it is exposed for pull streams.

I also add a `jetstream_push.rs` example (mostly copied from `jetstream_pull.rs`) which I used to confirm that it works, but I am happy to remove this if it is unhelpful.

I changed the dosctring on the pull consumer from “Threshold for *ephemeral* consumer activity” to “Threshold for consumer activity” to account for the fact that as of 2.9 it applies to non-ephemeral consumers also.


Jarema (approved):
LGTM!

PR #1000, opened 2023-06-16
Fix KV update so that it works properly with a JS domain

protochron:
It looks like the code for calling update on a KV bucket doesn't include the JS domain on it, so it gives you an error when attempting to make that call.


Jarema (approved):
Thanks for your contribution and the fix!

LGTM!

PR #1004, opened 2023-06-25
Make the current `Error` type work with anyhow

paolobarbolini:
This makes the `Error` type work with `anyhow` (and I guess a few more crates) that [need the error to be `'static`](https://docs.rs/anyhow/1.0.71/anyhow/struct.Error.html#impl-AsRef%3Cdyn+Error+%2B+Sync+%2B+Send+%2B+'static%3E-for-Error)


caspervonb (approved):

Jarema (approved):
LGTM!
Thanks for contribution @paolobarbolini  :)

PR #1013, opened 2023-07-05
Bump rustls-webpki to v0.101.1

paolobarbolini:
Today rustls 0.21.3 came out and it upgraded to rustls-webpki v0.101. They can do this because they don't expose the dependency in their public API, so it's not a breaking change for them to do. This means though that now we have to compile two different versions of the same dependency. This fixes it by realizing the rustls-webpki versions.

### DIFF
--- a/async-nats/Cargo.toml
+++ b/async-nats/Cargo.toml
@@ -39,7 +39,7 @@ base64 = "0.21"
 tokio-retry = "0.3"
 ring = "0.16"
 rand = "0.8"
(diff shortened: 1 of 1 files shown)


paolobarbolini:
Let's wait the resolution of https://github.com/rustls/webpki/issues/112 before merging

paolobarbolini:
This can now be merged

Jarema (approved):
LGTM!

Thanks for catching it @paolobarbolini

Issue #717, opened 2022-11-11
Implement `Deserialize` for `ServerAddr`


andrewhalle:
#### Use Case:

Including NATS connection information in a config file which is deserialized
with `serde`.

#### Proposed Change:

Derive (or otherwise implement) `Deserialize` on `ServerAddr` (possibly behind
a `serde` feature).

#### Who Benefits From The Change(s)?

Anyone using `serde`. Right now, we deserialize the connection URL as a
`String` in our configuration struct, and handle errors due to invalid NATS
URLs separately. It would be nice if we could make invalid NATS URLs
unrepresentable and handle errors as normal deserialization errors.

Jarema:
Stale issue. I thinks its quite easy to do it on user-end.

PR #1022, opened 2023-07-11
Implement Deserialize trait for ServerAddress

nepalez:
See: https://github.com/nats-io/nats.rs/issues/717

Impement the serde deserializer for the ServerAddress(Url) struct based on the existing implementation of `from_str`.


caspervonb:
I'm not sure about this, as there are a lot of types that are serialized with try_from/from_str via serde itself. @Jarema wdyt?

PR #1024, opened 2023-07-11
Mark `pull_sequence` as a slow test

n1ghtmare:
This tests takes 40 minutes on my Linux machine and is very slow on my mac (M2) as well.


caspervonb (approved):
lgtm

PR #1032, opened 2023-07-17
Fix unwrap from `HeaderName::from_str` call

### LINKED ISSUE #1030: Alphanumeric enforcement on `HeaderName` breaks Implementation with Memphis.
As stated above the Alphanumeric enforcement introduced in 0.30.0 breaks custom headers used by memphis.
https://github.com/nats-io/nats.rs/blob/cb0e15a7bad7a910266628662c1a477e46c7c106/async-nats/src/header.rs#L502-L504

Memphis uses Nats in the background and uses SpecialHeaders like `$memphis_producedBy` which inclue `$`, `_` and therefor triggering the `ParseHeaderNameError`

It also appears that the enforcement dosn't happen while inserting into the HeaderMap but rather when reading it at https://github.com/nats-io/nats.rs/blob/cb0e15a7bad7a910266628662c1a477e46c7c106/async-nats/src/connection.rs#L295
Causing the entire library to panic rather than returning an error.

But I don't understand why the headers are being restricted in the first place, as something like the headers above work absolutly fine, and I didn't find anything in the NATS documentation restricting the use of headers like that. I think if a user wants to use funky HeaderNames just let them do it.

caspervonb:
Closes #1030

### DIFF
--- a/async-nats/src/connection.rs
+++ b/async-nats/src/connection.rs
@@ -292,7 +292,10 @@ impl Connection {
                     value.push_str(v);
                 }
 
-                headers.append(HeaderName::from_str(key).unwrap(), value.trim().to_string());
+                let name = HeaderName::from_str(key)
+                    .map_err(|err| io::Error::new(io::ErrorKind::InvalidInput, err))?;
+
+                headers.append(name, value.trim().to_string());
             }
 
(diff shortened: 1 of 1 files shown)


Jarema (approved):
LGTM!

PR #1035, opened 2023-07-17
Add missing sequence number reset

paolobarbolini:
Doesn't fix my problem but it's something I noticed. The sequence number gets reset below but not here


Jarema (approved):
Good catch!
Thanks.

LGTM.

caspervonb (approved):
lgtm

PR #1039, opened 2023-07-17
Remove memcpy in object store PUT

paolobarbolini:
Removes a memcpy from the object store PUT implementation and a [scary](https://github.com/rust-lang/rust/issues/53827) `Box::new` with a huge array


caspervonb (approved):
LGTM!

PR #1040, opened 2023-07-17
Optimize `GET`ting and object's payload

paolobarbolini:
Removes a memcpy and an allocation from the object read implementation

### DIFF
--- a/async-nats/src/jetstream/object_store/mod.rs
+++ b/async-nats/src/jetstream/object_store/mod.rs
@@ -12,6 +12,7 @@
 // limitations under the License.
 
 //! Object Store module
+use std::collections::VecDeque;
 use std::fmt::Display;
 use std::{cmp, str::FromStr, task::Poll, time::Duration};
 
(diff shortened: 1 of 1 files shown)


async-nats/src/jetstream/object_store/mod.rs
```diff
@@ -568,7 +569,7 @@ impl Stream for List<'_> {
 /// Represents an object stored in a bucket.
 pub struct Object<'a> {
     pub info: ObjectInfo,
-    remaining_bytes: Vec<u8>,
+    remaining_bytes: VecDeque<u8>,
```
caspervonb, line 572: What do you think about using a bytes buffer here instead?

caspervonb (commented):
Was wondering if we could get away with having remaiing_bytes as a `BytesMut` buffer here.

paolobarbolini:
> Was wondering if we could get away with having remaiing_bytes as a `BytesMut` buffer here.

It could work but if you don't need a `Bytes` at the end you're just wasting time with the ref-counting and all of the internal `bytes` machinery

caspervonb (approved):
lgtm

Jarema (approved):
LGTM!

PR #1051, opened 2023-07-22
Remove useless use of `Option` around `HeaderMap`

paolobarbolini:
Calling `HeaderMap::new` doesn't allocate, but instead creates an [empty hashmap](https://doc.rust-lang.org/nightly/std/collections/hash_map/struct.HashMap.html#method.new).

Considering that code already has to check whether there really were any headers inside of the `HeaderMap` when the variant is `Some`, the `Option` around it didn't make sense.

### DIFF
--- a/async-nats/src/client.rs
+++ b/async-nats/src/client.rs
@@ -152,15 +152,8 @@ impl Client {
     /// # }
     /// ```
     pub async fn publish(&self, subject: Subject, payload: Bytes) -> Result<(), PublishError> {
-        self.sender
-            .send(Command::Publish {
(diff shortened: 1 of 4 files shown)


caspervonb:
> Considering that code already has to check whether there really were any headers inside of the HeaderMap when the variant is Some, the Option around it didn't make sense.

Just a bit of trivia as to why, before we used to send HMSG if headers were Some, even if they were empty.

caspervonb:
Boils down to this I think, Option signals semantics. Is an empty header map lighter than an empty header map?

PR #1059, opened 2023-07-28
Clearer documentation for Context::get_object_store

nmandery:
This is only a minor change which makes the comment on `Context::get_object_store` a bit clearer - the former comment was the same as on `create_object_store`.


Jarema (approved):
LGTM!
Thanks for your contribution.

PR #1073, opened 2023-08-02
Remove some allocations from benchmarks

paolobarbolini:
Makes the allocator work less during benchmarks, reducing allocation noise and other unwanted side-effects


caspervonb:
Needs a rebase.

caspervonb (approved):

PR #1089, opened 2023-08-14
Refactor ParseError

nepalez:
This PR is a part of error specialization initiated by the https://github.com/nats-io/nats.rs/pull/1047

See the initial discussion of the change in this https://github.com/nats-io/nats.rs/pull/1081 (to be removed later, after processing the feedback by @Jarema and creating smaller PR-s from its commits).

This particular PR is created to follow-up [the discussion](https://github.com/nats-io/nats.rs/pull/1081/files#r1288033272)

---

In this PR instead of the `crate::Error`, the <private> method `stream::parse_headers` will return the specialized error with kind enumerating the possible causes of parsing errors. This change will help to specialize errors error returned by public methods.

Along this change, I also refactored 2 errors, namely the `ParseHeaderNameError` and the `ParseHeaderValueError` to convert them to specialization of generic `crate::error::Error<_>`.

These changes are collected in the same commit because all of them describe errors occurred during either building or parsing headers.

### DIFF
--- a/async-nats/src/header.rs
+++ b/async-nats/src/header.rs
@@ -22,6 +22,7 @@
 
 use std::{collections::HashMap, fmt, slice::Iter, str::FromStr};
 
+use crate::error::Error;
 use bytes::Bytes;
 use serde::{Deserialize, Serialize};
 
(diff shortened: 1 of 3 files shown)


async-nats/src/header.rs
```diff
@@ -339,19 +340,26 @@ impl HeaderValue {
     }
 }
 
-#[derive(Debug, Clone)]
-pub struct ParseHeaderValueError;
```
caspervonb, line 343: These are intended to be discrete simple errors for `FromStr`, they live outside of the "io-esque" methods on context and client.

This complicates what is a very simple and straight forward discrete error type imo.

PR #1100, opened 2023-08-29
fix/doc: Typo in ping_interval example

sravan-s:
Issue: ping_interval example was calling flush_interval
Also, Use duration::from_secs


Jarema (approved):
LGTM!
Thanks for the contribution!

PR #1106, opened 2023-09-13
Switch dependency: json->serde_json in nats-server

nepalez:
This patch fixes the [cargo-deny](https://github.com/EmbarkStudios/cargo-deny)'s security warning for `nats-server` which is used by `async-nats`:

```shell
$ cargo-deny --exclude nats --exclude nats_test_server check
```

which reported:

```
   ┌─ ../nats.rs/Cargo.lock:83:1
   │
83 │ json 0.12.4 registry+https://github.com/rust-lang/crates.io-index
   │ ----------------------------------------------------------------- unmaintained advisory detected
   │
   = ID: RUSTSEC-2022-0081
   = Advisory: https://rustsec.org/advisories/RUSTSEC-2022-0081
   = Last release was almost 3 years ago.

   The maintainer is unresponsive with outstanding issues.
```

After the fix `cargo-deny` got satisfied

```
advisories ok, bans ok, licenses ok, sources ok
```

### DIFF
--- a/nats-server/src/lib.rs
+++ b/nats-server/src/lib.rs
@@ -22,6 +22,7 @@ use std::{thread, time::Duration};
 use lazy_static::lazy_static;
 use rand::Rng;
 use regex::Regex;
+use serde_json::{self, Value};
 
 pub struct Server {
     inner: Inner,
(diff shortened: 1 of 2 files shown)


nats-server/src/lib.rs
```diff
@@ -22,6 +22,8 @@ use std::{thread, time::Duration};
 use lazy_static::lazy_static;
 use rand::Rng;
 use regex::Regex;
+use serde_json as json;
```
caspervonb, line 25: Might as-well qualify this.

```suggestion
use serde_json;
```

caspervonb (commented):
Tiny nit

nats-server/src/lib.rs
nepalez: Fixed along with a friendlier error message in case the parsing failed.

caspervonb (approved):

PR #1108, opened 2023-09-15
Fix minimal versions

nepalez:
Here I checked and then fixed minimal versions for transitive dependencies of the `async-nats`.

```shell
cargo +nightly update -Zminimal-versions
cargo check --package async-nats --all-targets
```

After the fixes it got green.

See detailed explanations in the commit messages


Jarema (approved):
LGTM!

Thanks for your contribution!

PR #1123, opened 2023-09-29
Always log `Event`s

paolobarbolini:
This removes an allocation from the `Box`ing of the `Future` returned by the event handler closure and makes it always trace the event, which makes much more sense to me.


Jarema (approved):
Yes, I agree, that makes sense.

LGTM!

PR #1139, opened 2023-10-14
Changed `self` to a borrow in `Request::respond()`

jlandahl:
This addresses https://github.com/nats-io/nats.rs/issues/1138


Jarema (approved):
Hey.
We agreed that it will be useful to have tor service that send multiple responses.

LGTM!

PR #1152, opened 2023-11-02
Calculate average processing time as total time / # requests

jadamcrain:
Proposed fix for https://github.com/nats-io/nats.rs/issues/1151.

I wasn't crazy about any of the options for dividing a `Duration`.

The problem w/ [checked_div](https://doc.rust-lang.org/stable/std/time/struct.Duration.html#method.checked_div) is that it only takes `u32` and total_requests is `usize`.

I considered div_f64, but just doing the calculation as nanos and converting back to a Duration seems fool-proof and panic-free.


Jarema (approved):
LGTM! 

PR #1161, opened 2023-11-14
feat(subject): Adds `into_string` method to `Subject`

thomastaylor312:
This should allow for easier conversion back into a `String`, possibly without an allocation if the underlying data is already owned


Jarema (approved):
LGTM!

PR #1164, opened 2023-11-28
Remove unused dependency on http

oscarwcl:
This was only used for generating a documentation link. I've changed it to not link to the `http::header` docs at all, but this could also be made a manual link to docs.rs if that's preferred.


Jarema (approved):
LGTM!

Thanks for the contribution!

PR #1167, opened 2023-11-30
Don't expose BoxFuture in public API

oscarwcl:
Based on #1166 

`ObjectStore::get` calls itself recursively, so the returned future needs to be boxed. However, this is an implementation detail, there's no reason to expose it in the public API. Instead, we can move the recursive call to a separate `get_impl` function that handles the boxing, and `get` can just be a standard async function.

This is a minor stylistic change, but if it's desired it makes sense to land it at the same time as #1166 since that is already a breaking change.


Jarema (approved):
LGTM!

PR #1191, opened 2024-01-06
Do asynchronous DNS lookups in `ServerAddr::socket_addrs`

paolobarbolini:
Turns out `ServerAddr::socket_addrs` was blocking the tokio runtime because it used the std blocking implementation for doing DNS lookups. This switches to the `tokio` implementation, which internally does a `spawn_blocking` of the std call.


Jarema (approved):
LGTM!

Thank you for catching and fixing this @paolobarbolini !

PR #1199, opened 2024-01-25
fix: ensure the ping interval waker is registered

dodomorandi:
When an interval is ready, it does not need to register its internal waker, leading to the future not being polled anymore if the connection received no data.

The fix is pretty simple, and it should never cause any problems unless the interval duration is very close to zero, which will probably cause strange issues anyway.


Jarema (approved):
LGTM!

Thanks for catching this one!

PR #1207, opened 2024-02-03
Make `Subject::from_static` a const fn

paolobarbolini:
The whole point of having a `from_static` function is to use it in a const context :wink:


Jarema (approved):
good catch :)

LGTM!

PR #1223, opened 2024-03-02
Bump base64 to v0.22

paolobarbolini:
`base64` is one of those dependencies that get used by everybody and ends up finding itself into projects in the form of multiple duplicate versions. Since you're about to make a release let's do our part in not having duplicate versions of it :smiley:.

Changelog: https://github.com/marshallpierce/rust-base64/blob/master/RELEASE-NOTES.md#0220


Jarema (approved):
LGTM! :)

PR #1231, opened 2024-03-08
Add `ToServerAddrs` impl for array/vector of strings

mmalek:
Quality-of-life improvement for `connect()` function. With this changes it's straightforward to pass `Vec<String>` or `[&str]` directly into `connect()` function.
This is useful when passing list of NATS servers directly from configuration file or command-line argument.


Jarema (approved):
LGTM!

Thanks for your contribution!

PR #1236, opened 2024-03-15
feat(client): Wraps inbox prefix in an `Arc`

thomastaylor312:
The goal of the `Client` struct was to be a clonable handle. However, the struct contained a bare string that was cloned each time. In apps that clone the client regularly, this lead to quite a few allocations.

This is a non-breaking change purely internal to the client

### DIFF
--- a/async-nats/src/client.rs
+++ b/async-nats/src/client.rs
@@ -69,7 +69,7 @@ pub struct Client {
     pub(crate) sender: mpsc::Sender<Command>,
     next_subscription_id: Arc<AtomicU64>,
     subscription_capacity: usize,
-    inbox_prefix: String,
+    inbox_prefix: Arc<str>,
     request_timeout: Option<Duration>,
     max_payload: Arc<AtomicUsize>,
 }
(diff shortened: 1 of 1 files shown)


async-nats/src/client.rs
```diff
@@ -69,7 +69,7 @@ pub struct Client {
     pub(crate) sender: mpsc::Sender<Command>,
     next_subscription_id: Arc<AtomicU64>,
     subscription_capacity: usize,
-    inbox_prefix: String,
+    inbox_prefix: Arc<String>,
```
rvolosatovs, line 72:
```suggestion
    inbox_prefix: Arc<str>,
```

`Arc<str>` is generally a more appropriate type for this use case

async-nats/src/client.rs
```diff
@@ -90,7 +90,7 @@ impl Client {
             sender,
             next_subscription_id: Arc::new(AtomicU64::new(1)),
             subscription_capacity: capacity,
-            inbox_prefix,
+            inbox_prefix: Arc::new(inbox_prefix),
```
rvolosatovs, line 93:
```suggestion
            inbox_prefix: inbox_prefix.into(),
```
Is probably the way to go here

rvolosatovs (approved):

Jarema (approved):
LGTM!

Thanks for the contribution!

PR #1241, opened 2024-04-05
Don't force flush if write buffer isn't empty

paolobarbolini:
It turns out `poll_write` could return `Poll::Pending` while `poll_flush` could flush successfully (`Poll::Ready`). This seems like a rare case, although possible in theory.  This fixes it by actually making it behave the way it was described in the original PR https://github.com/nats-io/nats.rs/pull/1060#issuecomment-1657242359


Jarema (approved):
LGTM!

PR #1258, opened 2024-05-10
Preserve case of server error messages

oscarwcl:
Server errors can contain case-sensitive values like subject names which can cause confusion if not displayed in their original case.

This was motivated by this error: `nats: permissions violation for publish to "$js.api.stream.info.kv_test"`, which is misleading since the actual subject is `$JS.API.STREAM.INFO.KV_test`.

### DIFF
--- a/async-nats/src/lib.rs
+++ b/async-nats/src/lib.rs
@@ -1294,7 +1294,8 @@ impl ServerError {
     fn new(error: String) -> ServerError {
         match error.to_lowercase().as_str() {
             "authorization violation" => ServerError::AuthorizationViolation,
-            other => ServerError::Other(other.to_string()),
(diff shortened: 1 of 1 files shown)


caspervonb (approved):
lgtm

async-nats/src/lib.rs
```diff
@@ -1294,7 +1294,7 @@ impl ServerError {
     fn new(error: String) -> ServerError {
         match error.to_lowercase().as_str() {
             "authorization violation" => ServerError::AuthorizationViolation,
-            other => ServerError::Other(other.to_string()),
+            _ => ServerError::Other(error),
```
caspervonb, line 1297: Nit, add a little comment calling out it's done to preserve casing?

PR #1278, opened 2024-06-17
Allow setting size limit for object storage buckets

liamkinne:
In line with the Go client: https://github.com/nats-io/nats.go/blob/370bc4fc4cac2de8d5dbbddbb98739f9cb2d7f8b/object.go#L283


Jarema (approved):
LGTM!

PR #1300, opened 2024-08-07
De/Serialize `sample_frequency` correctly for Push and Pull Consumers

bengsparks:
Closes #1299 

This PR instructs `serde` to rename  `sample_frequency` into `sample_freq` during de/serialization.
It also introduces a module for converting the `u8` into a `String` [due to nats-server's handling of the SampleFrequency field](https://github.com/nats-io/nats-server/blob/d1daf5f7916f088b8ea65c6694ab9b3074f8187b/server/consumer.go#L80).

The PR is split into two separate commits, one for `async-nats` and the other for `nats`, so that the latter can be reverted if it is not wanted.

This PR does not contain any tests, but used the setup from #1299 to check if the field was being correctly set.

PR #1303, opened 2024-08-21
nats: De/Ser using `sample_freq` instead of `sample_frequency`

bengsparks:
Addendum to #1300 for the `nats` module.

This PR instructs serde to rename `sample_frequency` into `sample_freq` during de/serialization.
It also introduces a module for converting the `u8` into a `String`.

Finally, this PR adds a test in `nats/tests/jetstream.rs` that can be executed via `cargo test -- jetstream_consumer_configs_sample_frequency`, which checks that the stored config of the consumer now contains the correctly set field.
Removing the `rename = sample_freq` directive from the `ConsumerConfig` type causes this test to fail, so all seems to be correct.


Jarema (approved):
LGTM!

Thanks for yet another contribution!

PR #1356, opened 2024-12-30
async-nats: check push consumers in `sample_freq` de/serialisation test

bengsparks:
Followup from #1354 

Previously merged test accidentally tested pull consumer configs twice, neglecting to test push consumer configs as well.


Jarema (approved):
LGTM

PR #1382, opened 2025-03-13
Verideath investigation

nickgerace:
Based on https://github.com/nats-io/nats.rs/commit/0eca616f20cf991a15630c189962b89addee0fb0 from upstream


nickgerace:
Oops I am sorry @Jarema!! This was supposed to be in our repository only 🤦‍♂️.

We will close this as soon as the investigation into a bug is done, if that is okay with you. Otherwise, we can close now

PR #1416, opened 2025-05-31
Fix the order of the max payload exceeded message numbers

fnichol:
Hi there!

While debugging a NATS-driven application we hit the "max message payload exceeded" error in the Rust client. It took a little re-reading to realize that the 2 numbers (that is, the `max_payload` limit and the `payload.len()` of the current message) were revered in the Rust error. This change swaps them 😄 

Thanks!


Jarema (approved):
LGTM!

PR #1422, opened 2025-06-14
Allow conversion from &str in ToSubject.

marknefedov:
I haven't discovered why can't we pass a borrowed str and it make api a bit more versatile.


Jarema (changes requested):
There are many ways you can achieve your goal.

This code does not compile, and will not compile - you cannot have overlapping implementations (in this case - overlapping with `&'static str`).

PR #1428, opened 2025-07-26
Expose get_info for consumer

nitsky:
This PR exposes the `get_info` method on consumer, just like it is for stream already.


Jarema (changes requested):
Thanks for the PR!

Please add docs (explaining the difference between `get_info` and `info`) with an example.

nitsky:
I added a comment copied from `get_info` for Stream, and a simple example.

Issue #1481, opened 2025-10-31
Exposing client from jetstream context


Ferrany1:
### Proposed change

Exposing client from jetstream context

### Use case

Since jetstream async_nats::jetstream::new consumes connection it is needed to create new/clone previous to check for connection state or implement graceful shutdown

But if for the ping it is acceptable, for drain its not, documentation says drain will prevent any new messages from being published

But if i will call drain on original connection (not the cloned one for a jetstream) jetstream will be still able to publish messages, so method

``
    pub const fn client(&self) -> &Client {
        &self.client
    }
``

might be useful for both usecases

### Contribution

Yes

Jarema:
Agree that this is not clear. It should state `client connection` or even simply `connection`.

As per docs:

```
/// Client is a `Cloneable` handle to NATS connection.
```

And it's nothing more. There is no `Connection` attached to it. `Cloning` it merely creates a new `handle clone`, not a new `conneciton`.

Will improve the docs on that topic unless you're interested in contributing.

EDIT: runnign it might fail because connection was not yet closed.

PR #1482, opened 2025-10-31
Feature: expose jetstream client by reference

Ferrany1:
#1481 1481


caspervonb:
Not sure what the motivation is, since client is clonable?

Could have:

```
fn client_ref(&self) -> &Client;
```

But would have to be tied to lifetime of self.

PR #1496, opened 2025-12-29
Upgrade dependency for `async-nats` for `nats-server`.

vkolomeyko:
Without this change during Cargo sync step, the following error is produced:
```
\nats.rs\nats-server> cargo.exe check --workspace --all-targets --features async-nats/default,async-nats/ring,async-nats/server_2_10,async-nats/server_2_11,async-nats/server_2_12,async-nats/service --keep-going
error: failed to select a version for `async-nats`.
    ... required by package `nats-server v0.1.0 (\nats.rs\nats-server)`
    ... which satisfies path dependency `nats-server` (locked to 0.1.0) of package `nats v0.25.0 (\nats.rs\nats)`
versions that meet the requirements `^0.38` (locked to 0.38.0) are: 0.38.0

package `nats-server` depends on `async-nats` with feature `server_2_11` but `async-nats` does not have that feature.
 package `async-nats` does have feature `server_2_10`


failed to select a version for `async-nats` which could resolve this conflict
```

Also, enabled `nats-server` tests on Windows. I have checked and they are passing locally.

### DIFF
--- a/nats-server/Cargo.toml
+++ b/nats-server/Cargo.toml
@@ -16,6 +16,6 @@ rand = "0.8"
 tokio-retry = "0.3.0"
 
 [dev-dependencies]
-async-nats = "0.45.0"
+async-nats = { path = "../async-nats" }
 tokio = { version = "1", features = ["full"]  }
 futures-util = "0.3"


vkolomeyko:
@Jarema

Jarema:
That's not a proper solution, as it makes all tests fail.

Can you provide a command that leads to this error?

vkolomeyko:
@Jarema - many thanks for approving. Unfortunately, I do not have permissions to merge this PR. Would you be able to merge it, please?

PR #1497, opened 2025-12-29
Fix error deserialization when err_code is absent

Jarema:
In rare cases nats-server returns JetStream error without err_code. This causes deserialization error that does not allow to peek into rest of given error information.


piotrpio (approved):
LGTM!

PR #1500, opened 2026-01-05
Fix race condition in nats-server wrapper utility for cluster creation

Jarema:
The tests were not using proper ports in case of reassignment. This commit also makes sure that GitHub CI is running the test.


piotrpio (approved):
LGTM

PR #1503, opened 2026-01-12
fix: properly wake self after drain is scheduled

OliverNChalk:
Without this fix the following sad situation happens:

- Client sends UNSUB message to server.
- Server unsubscribes client without ACK (idk why) or closing TCP (because other subs might exist).
- Client receives no further commands from the subscription holder.
- Server sends  no more messages.

The result is the worker thread idles for 60s before the next ping when it finally drops the sub letting the drain() caller shutdown.

## Fix

- Have the future tell the runtime it needs to immediately be rescheduled.

## Alternatives Considered

- `Option<Sender>` that set to None when drain is called.
  - Doesn't worker, Sender is sahred by all subs. Also would be annoying to have an option.

### DIFF
--- a/async-nats/tests/client_tests.rs
+++ b/async-nats/tests/client_tests.rs
@@ -1095,6 +1095,9 @@ mod client {
         client.publish("test", "data".into()).await.unwrap();
         client.flush().await.unwrap();
 
+        // TODO: Use proper synchronization, this is a horrible hack.
+        tokio::time::sleep(Duration::from_millis(100)).await;
+
         // Send the drain command
         sub.drain().await.expect("Expected to drain the sub");
 


caspervonb (approved):
lgtm

Jarema (changes requested):
acually - I found a bug that causes the tests to fail. 
We need to check for `Drain` with `sid` set to `None`. Otherwise we break behavior of breaking the `Drain` of a subsription.

cc @OliverNChalk

async-nats/src/lib.rs
```diff
@@ -583,6 +584,7 @@ impl ConnectionHandler {
                                 made_progress = true;
 
                                 for cmd in recv_buf.drain(..) {
+                                    wake_self |= matches!(&cmd, Command::Drain { .. });
```
Jarema:
```suggestion
                                    wake_self |= matches!(&cmd, Command::Drain { sid: None });
```

OliverNChalk: Pushed the fix for you to rerun CI: cdc073dbcc6ce3fd52aacd009a2a7595f47bbea6

Would ideally get a proper fix though...

Jarema: No worries. Thanks for helping out. This is much appreciated!

PR #1504, opened 2026-01-15
Fix slow subscription shutdown after drain

Jarema:
In cases when there was no traffic from server and `drain` on subscription was called, while the `drain` itself would be effective immediately, the `Subscriber` iterator could stay alive until next `Pong` from the server.
This commit ensures that:
1. After every `drain`, `Ping` will be dispatched
2. Subscriptions will be closed when `Pong` is received.

supersedes #1503 

### DIFF
--- a/async-nats/src/lib.rs
+++ b/async-nats/src/lib.rs
@@ -203,6 +203,7 @@ use tracing::{debug, error};
 
 use core::fmt;
 use std::collections::HashMap;
+use std::collections::VecDeque;
 use std::fmt::Display;
 use std::future::Future;
 use std::iter;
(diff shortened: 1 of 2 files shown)


Jarema:
cc @OliverNChalk take a look. This is the approach close to what Go client does.

async-nats/src/lib.rs
```diff
@@ -826,6 +831,7 @@ impl ConnectionHandler {
             Command::Drain { sid } => {
                 let mut drain_sub = |sid: u64, sub: &mut Subscription| {
                     sub.is_draining = true;
```
OliverNChalk: Is sub.is_draining ever read after this change?

Jarema: yup. Removed.

caspervonb (approved):
Tiny nit

PR #1510, opened 2026-01-27
Add subject validation

Jarema:


caspervonb (approved):
Ideally, we'd just have `Subject` imo, and you'd skip validation by doing Subject::new_unchecked() for example. This is what I have in mind for Python as-well.

But, we have backwards compatibility to consider, so maybe that's the future?

LGTM on the overall, Copilot has some nits.

PR #1514, opened 2026-02-14
Throw error instead of panic when reply subject is empty

### LINKED ISSUE #1511: Publish to service endpoint without reply subject causes panic
### Observed behavior

Publishing to a subject used by a service endpoint causes the service to panic when running `.respond()`.
See: https://github.com/nats-io/nats.rs/blob/4417346a772c365fe8f7b31d595c61dce190a007/async-nats/src/service/mod.rs#L757C9-L757C57

### Expected behavior

The `respond` call should return an error to be handled by the caller.

### Server and client version

NATS server 2.10
async-nats 0.45.0

### Host environment

N/A

### Steps to reproduce

1. Run any example service that calls `.respond()`
2. Publish without a reply subject specified: `nats pub test ""`

liamkinne:
Fixes #1511

### DIFF
--- a/async-nats/tests/service_tests.rs
+++ b/async-nats/tests/service_tests.rs
@@ -13,9 +13,11 @@
 
 #[cfg(feature = "service")]
 mod service {
+    use async_nats::client::PublishErrorKind;
     use async_nats::service::{self, Info, ServiceExt, Stats};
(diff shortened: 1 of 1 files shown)


Jarema (approved):
LGTM!

PR #1531, opened 2026-02-27
Remove sync from claude workflow

Jarema:
This caused re-review on every PR push.


piotrpio (approved):
LGTM

Issue #1526, opened 2026-02-27
`connection_timeout` only covers TCP connect, not the full NATS handshake


kfollesdal:
### Observed behavior

Hi, is this a bug or intentional? I am no rust expert, but found this observation while doing some debugging with help of ai. 

**Summary**

`ConnectOptions::connection_timeout()` wraps only `TcpStream::connect()` (the TCP SYN/SYN-ACK), but the subsequent NATS protocol handshake — reading `INFO`, writing `CONNECT+PING`, reading `PONG` — has no timeout at all. If a host is reachable at the TCP level but the NATS process is frozen or unresponsive, `connect() `hangs indefinitely.

**Where the gap is**

In `connector.rs`, the timeout is applied around `TcpStream::connect()`:

`tokio::time::timeout(self.connect_timeout, TcpStream::connect(addr)).await??;`

Once TCP connects successfully, the handshake calls (`read_op(), easy_write_and_flush()`, etc.) are completely unguarded. The ping/keepalive mechanisms only activate after the connection handler spawns, which happens after the handshake completes. The WebSocket paths have the same issue — timeout only wraps the WS/WSS connect, not the subsequent protocol exchange.

**How Go's NATS client handles this**

The Go client in nats.go solves this cleanly in processConnectInit() by setting a single deadline that covers the entire handshake:

```go
func (nc *Conn) processConnectInit() error {
    nc.conn.SetDeadline(time.Now().Add(nc.Opts.Timeout))
    defer nc.conn.SetDeadline(time.Time{})
    // ... reads INFO, does TLS upgrade, writes CONNECT+PING, reads PONG
    // All covered by the deadline above
}
```

This means Go's Timeout option covers:

✅ TCP connect
✅ Reading INFO
✅ TLS upgrade
✅ Writing CONNECT+PING
✅ Reading PONG

Whereas async-nats currently covers:

✅ TCP connect
❌ Reading INFO
❌ TLS upgrade
❌ Writing CONNECT+PING
❌ Reading PONG

**Suggested fix**

One approach (mirroring Go) would be to wrap the entire handshake sequence in a single tokio::time::timeout() using the existing connect_timeout value, rather than only the TCP connect. Something like:

```rust
tokio::time::timeout(self.connect_timeout, async {
    let stream = TcpStream::connect(addr).await?;
    // ... read INFO, TLS, write CONNECT+PING, read PONG
    Ok(stream)
}).await??;
```

**Workaround**

[...]

PR #1544, opened 2026-03-17
Fix connection_timeout to cover full NATS handshake

rahsonaw:
Previously, connection_timeout only guarded the TCP connect (SYN/ACK). The subsequent NATS protocol handshake steps — reading INFO, TLS upgrade, writing CONNECT+PING, and reading PONG — had no timeout, allowing the client to hang indefinitely if a server was reachable at the TCP level but unresponsive at the protocol level.

This uses a single deadline per connection attempt, computed once as Instant::now() + connection_timeout. Every I/O step consumes from the same budget via deadline.saturating_duration_since(Instant::now()), mirroring the Go client's approach in processConnectInit().

Fixes: https://github.com/nats-io/nats.rs/issues/1526

### DIFF
--- a/async-nats/src/connector.rs
+++ b/async-nats/src/connector.rs
@@ -208,31 +208,27 @@ impl Connector {
                         return Ok((server_info, connection));
                     }
 
-                    Ok(Err(inner)) => {
-                        match inner.kind() {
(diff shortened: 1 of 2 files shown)


Jarema (changes requested):
Did you consider the single timeout approach?

PR #1551, opened 2026-03-29
Add Client::rtt for async-nats

### LINKED ISSUE #909: Implement `Client::rtt` for async_nats

MattPatchava:
Implement Client::rtt() which measures the round trip time to the server by sending a PING and awaiting the corresponding PONG.

Closes #909

Implements Client::rtt() for async-nats. Each RTT request sends its own PING with a per-ping timestamp and oneshot sender stored in a VecDeque, replacing the previous usize pending_pings counter to correctly correlate PING/PONG pairs even with concurrent callers and interleaved keepalive pings.

### DIFF
--- a/.config/nats.dic
+++ b/.config/nats.dic
@@ -207,3 +207,5 @@ LF
 nats
 WebSocket
 rng
+rtt
+RTT
--- a/async-nats/src/client.rs
+++ b/async-nats/src/client.rs
@@ -799,6 +799,35 @@ impl Client {
         Ok(())
     }
 
+    /// Calculates the round trip time between this client and the server by sending a PING and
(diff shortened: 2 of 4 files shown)


Jarema (commented):
Thanks for the PR!

However, I wonder if this is not a bit of overcomplication.

Did you consider approach similar to how other clients address it with `waiter` queue?

MattPatchava:
Hey @Jarema, rebased onto main. PR is clean now, just the one commit. Could you take another look when you get a chance?

PR #1554, opened 2026-03-29
Add respond_with_headers method to handle responses with custom headers

### LINKED ISSUE #1553: Add new `respond_with_headers` fn to Nats Service Framework
### Proposed change

Add an API to `async-nats`'s `service::Request` that allows responding with custom headers while preserving the existing request statistics/timing behavior. Concretely, introduce a new method:

- `pub async fn respond_with_headers(&self, headers: HeaderMap, response: Result<Bytes, error::Error>) -> Result<(), PublishError>;`

and make the existing `respond` call into it (preserving backwards compatibility):

- `pub async fn respond(&self, response: Result<Bytes, error::Error>) -> Result<(), PublishError> { self.respond_with_headers(HeaderMap::new(), response).await }`

This keeps the existing internal stats accounting, error header behavior for the NATS `Nats-Service-Error` / `Nats-Service-Error-Code` fields, and allows frameworks to set additional headers (e.g. `x-success`, tracing, or semantic metadata) when replying.

### Use case

Frameworks built on top of `async-nats` need to attach and inspect protocol headers on replies for correct client-side behavior and better DX. For example:

- A typed RPC framework wants to set an explicit `x-success: true|false` header so generated clients can branch on success vs. failure without guessing payload shapes (avoiding brittle decode heuristics).
- A framework needs to attach tracing, feature flags, or authorization metadata to the reply headers so clients can read them without parsing payloads.
- The `async-nats` crate already supports sending with headers (`publish_with_headers`)—there is no reason the `Request` API shouldn't expose that capability.

This change preserves the current `respond` behaviour (which updates endpoint stats and sets `Nats-Service-Error` headers for error responses) while allowing callers to add arbitrary headers for richer protocols.

### Contribution

```rs
// In `src/service/mod.rs` (impl Request)
pub async fn respond(&self, response: Result<Bytes, error::Error>) -> Result<(), PublishError> {
    // Backwards-compatible: delegate to the new, header-capable method.
    self.respond_with_headers(HeaderMap::new(), response).await
}

/// Sends response for the request with additional headers.
///
/// This preserves the existing stats and error-handling semantics while
/// allowing callers to attach headers to the reply (e.g. `x-success`).
pub async fn respond_with_headers(
    &self,
    mut headers: HeaderMap,
    response: Result<Bytes, error::Error>,
) -> Result<(), PublishError> {
    let reply = self.message.reply.clone().unwrap();
    let result = match response {
        Ok(payload) => {
            // If callers provided headers, publish using the header-aware API.
            if headers.is_empty() {
                self.client.publish(reply, payload).await
            } else {
                self.client.publish_with_headers(reply, headers, payload).await
            }
        }
        Err(err) => {
            // Maintain existing stats bookkeeping on error responses.
            self.stats
                .lock()
                .unwrap()
                .endpoints
                .entry(self.endpoint.clone())
                .and_modify(|stats| {
                    stats.last_error = Some(err.clone());
                    stats.errors += 1;
                })
                .or_default();

            // Preserve the NATS service error headers
            headers.insert(NATS_SERVICE_ERROR, err.status.as_str());
            headers.insert(NATS_SERVICE_ERROR_CODE, err.code.to_string().as_str());

            self.client
                .publish_with_headers(reply, headers, "".into())
                .await
        }
    };

    // Keep the existing timing/stat tracking intact.
    let elapsed = self.issued.elapsed();
    let mut stats = self.stats.lock().unwrap();
    let stats = stats.endpoints.get_mut(self.endpoint.as_str()).unwrap();
    stats.requests += 1;
    stats.processing_time += elapsed;
    stats.average_processing_time = {
        let avg_nanos = (stats.processing_time.as_nanos() / stats.requests as u128) as u64;
        Duration::from_nanos(avg_nanos)
    };
    result
}
```

giddyos:
Closes #1553

### DIFF
--- a/async-nats/src/service/mod.rs
+++ b/async-nats/src/service/mod.rs
@@ -764,7 +764,7 @@ impl Request {
     /// On error responses, [Nats-Service-Error][NATS_SERVICE_ERROR] and
     /// [Nats-Service-Error-Code][NATS_SERVICE_ERROR_CODE] are always set from the provided
     /// [`error::Error`]. If the provided [HeaderMap] already contains values for either
-    /// of those headers, they will be overriden. All other user-supplied headers
+    /// of those headers, they will be overridden. All other user-supplied headers
     /// are preserved.
     ///
     /// # Examples


giddyos:
Made the parameter optional to avoid creating unused HeaderMap when the normal respond() method is called. Very straightforward changes with backwards compatibility and is required because service does not make its internal stats mutable so people using service framework have no way of adding headers to response without tossing away the internal stats keeping

giddyos:
updated documentation and added 2 new tests that prove headers function as expected

PR #1567, opened 2026-04-14
Add `try_server_info` and `max_payload` methods

yordis:
- retry_on_initial_connect can expose placeholder ServerInfo values before the first INFO frame arrives, which makes downstream clients treat max_payload as zero.
- callers need a way to distinguish between real server metadata and the pre-connect placeholder state without forcing a breaking API change.
- aligning the compatibility fallback with the existing 1 MiB internal default removes the current footgun while preserving existing callers.

### DIFF
--- a/async-nats/src/client.rs
+++ b/async-nats/src/client.rs
@@ -77,7 +77,7 @@ impl Display for PublishErrorKind {
 /// [crate::connect] and [crate::ConnectOptions::connect]
 #[derive(Clone, Debug)]
 pub struct Client {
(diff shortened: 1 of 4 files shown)


Jarema:
@yordis I think the way forward here is:

1. The new method is smart. 
2. Keep the old method in the original state. Do not pollute Server Info by client defaults.
3. add new method for checking current `max_payload`

yordis:
@Jarema check again

Jarema (approved):
LGTM!

Thanks for your contribution!

PR #1568, opened 2026-04-14
Remove needless async for service functions that don't require it

liamkinne:
These functions are actually fine as blocking.


Jarema:
Hey! 
Thanks for the PR!

Although those `async` keywords seems needless, they are very cheap and allow us to introduce changes that might require IO / other asynchronous calls inside them without introducing breaking changes to the public API.

Jarema:
Im closing this PR for now. Feel free to reopen if you think there is something more to discuss.

PR #1571, opened 2026-04-21
Bump to thiserror v2

aumetra:
Straightforward dependency upgrade. Eliminates the duplicate dependency in downstream projects.


Jarema (approved):
LGTM.

Thanks for the contribution. 
However - as I follow up, I would like to remove the dependency entirely.

PR #1573, opened 2026-04-23
Fix panic on reading non-utf8 data

### LINKED ISSUE #1572: Non UTF-8 data can cause `connect_with_options` to panic the background data ingestion thread
### Observed behavior

Panic with non UTF-8 data

### Expected behavior

Shouldn't panic with non UTF-8 data

### Server and client version

latest

### Host environment

_No response_

### Steps to reproduce

The fix is to remove the `.unwrap()`s. Write up:

[async_nats::ConnectOptions::connect](https://docs.rs/async-nats/0.45.0/src/async_nats/options.rs.html#171-173) calls [async_nats::connect_with_options](https://docs.rs/async-nats/0.45.0/src/async_nats/lib.rs.html#997-1087) which creates a [ConnectionHandler](https://docs.rs/async-nats/0.45.0/src/async_nats/lib.rs.html#437-448) which contains a Connection. The connection's poll_read_op method is called [here](https://docs.rs/async-nats/0.45.0/src/async_nats/lib.rs.html#537), which then calls try_read_op, which does not properly handle invalid UTF-8 data ([source](https://docs.rs/crate/async-nats/0.45.0/source/src/connection.rs#180)).

Original issue: https://github.com/vectordotdev/vector/issues/24399

Jarema:
nats-server does not valudate if subjects are valid UTF-8 so its possible to receive non-utf8 data.

fixes #1572

### DIFF
--- a/async-nats/src/connection.rs
+++ b/async-nats/src/connection.rs
@@ -177,7 +177,8 @@ impl Connection {
         }
 
         if self.read_buf.starts_with(b"MSG ") {
-            let line = str::from_utf8(&self.read_buf[4..len]).unwrap();
+            let line = str::from_utf8(&self.read_buf[4..len])
+                .map_err(|err| io::Error::new(io::ErrorKind::InvalidInput, err))?;
             let mut args = line.split(' ').filter(|s| !s.is_empty());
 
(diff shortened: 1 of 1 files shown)


caspervonb (approved):
lgtm

PR #1587, opened 2026-05-19
feat(header): add TryFrom impls for HeaderName

yordis:
- Library code should not panic on caller-supplied input. `HeaderName` had no standard fallible conversion, so panic-free callers either routed through `FromStr` (re-allocating any owned `String`) or duplicated the validation logic externally — two copies of the same rule that drift over time.

- `HeaderValue` is intentionally left untouched: the existing panicking `From<&str>` / `From<String>` impls collide with `core`'s blanket `impl<T, U> TryFrom<U> for T where U: Into<T>`, so a real `TryFrom` cannot coexist with them. `FromStr` remains the fallible path there until the panicking impls can be removed in a separate change.

### DIFF
--- a/async-nats/src/header.rs
+++ b/async-nats/src/header.rs
@@ -742,6 +742,7 @@ impl FromStr for HeaderName {
     }
 }
 
+/// Fallible conversion from a borrowed string slice; validates the name without panicking.
 impl TryFrom<&str> for HeaderName {
     type Error = ParseHeaderNameError;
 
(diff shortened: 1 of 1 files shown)


async-nats/src/header.rs
```diff
@@ -789,6 +818,17 @@ impl std::fmt::Display for ParseHeaderNameError {
 
 impl std::error::Error for ParseHeaderNameError {}
 
+fn validate_header_name(s: &str) -> Result<(), ParseHeaderNameError> {
+    // TODO: `c as u8` truncates non-ASCII codepoints to their low 8 bits, so chars
```
yordis: My intent here is to follow up on a different PR if we agree on it, I wasn't sure if I should fix it on this one, so I left behind the context

Jarema (changes requested):
Looks good. 
However I think the trunation but is worth fixing with one-liner fix: replace the `(c as u8)` predicate with `!c.is_ascii_graphic() || c == ':'`.

Jarema (approved):
LGTM!

PR #1588, opened 2026-05-24
feat(client): expose websocket headers via ConnectOptions

vladndd:
## Summary

Add `ConnectOptions::websocket_headers(http::HeaderMap)` so callers can inject custom HTTP headers on the WebSocket upgrade request when connecting to NATS through a reverse proxy that requires session-side validation.

## Motivation

A deployment proxying the NATS WebSocket through a reverse proxy commonly wants to validate a session `Cookie:` (or `Authorization:`) header on the WS upgrade before proxying the connection through to NATS. Today `Connector` builds its `tokio_websockets::client::Builder` with no caller hook between `.uri()` and `.connect()`, so there is no way to thread an upgrade-time header through.

`tokio-websockets` itself already exposes `Builder::add_header(name, value)`. This change plumbs one new field through `ConnectOptions → ConnectorOptions → connector::try_connect` and calls `add_header` in a loop before `.connect()`.

## API

\`\`\`rust
impl ConnectOptions {
    #[cfg(feature = "websockets")]
    pub fn websocket_headers(mut self, headers: http::HeaderMap) -> ConnectOptions;
}
\`\`\`

Default is an empty `HeaderMap`. Callers that don't touch the new method see byte-for-byte identical behaviour to upstream.

## Withdrawn

Closing this PR — withdrawing the contribution at my team's request before formal review. Apologies for the noise.

### DIFF
--- a/async-nats/Cargo.toml
+++ b/async-nats/Cargo.toml
@@ -42,6 +42,7 @@ rand = "0.10.1"
 rustls-webpki = { package = "rustls-webpki", default-features = false, version = "0.103.10" }
 portable-atomic = "1"
 tokio-websockets = { version = "0.10", default-features = false, features = ["client", "rand", "rustls-webpki-roots"], optional = true }
+http = { version = "1", default-features = false, features = ["std"], optional = true }
 pin-project = "1.0"
 
 [dev-dependencies]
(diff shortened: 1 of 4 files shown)


vladndd:
Closing — withdrawing this contribution at my team's request. Apologies for the noise; we'll re-evaluate the path forward.

PR #1589, opened 2026-05-25
feat(client): add websocket_headers to ConnectOptions

faylxen:
Allow callers to inject custom HTTP headers into the WebSocket upgrade
 request via ConnectOptions::websocket_headers(http::HeaderMap).

 This enables connecting to NATS through reverse proxies (Kong, nginx,
 Envoy) that validate session cookies or authorization tokens on the
 HTTP upgrade before proxying the WebSocket connection to NATS.

 The configured headers are reused on every connection and reconnection
 attempt. Non-WebSocket connections are unaffected.

### DIFF
--- a/async-nats/Cargo.toml
+++ b/async-nats/Cargo.toml
@@ -41,7 +41,8 @@ ring = { version = "0.17", optional = true }
 rand = "0.10.1"
 rustls-webpki = { package = "rustls-webpki", default-features = false, version = "0.103.10" }
 portable-atomic = "1"
-tokio-websockets = { version = "0.10", default-features = false, features = ["client", "rand", "rustls-webpki-roots"], optional = true }
+tokio-websockets = { version = "0.13", default-features = false, features = ["client", "rand", "rustls-webpki-roots"],  optional = true }
+http = { version = "1", default-features = false, features = ["std"], optional = true }
 pin-project = "1.0"
 
 [dev-dependencies]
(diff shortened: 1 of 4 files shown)


Jarema (changes requested):
Hey,

Thanks for the PR!

This is somethig we need for full websocket parity.

Few points to address:

  Parity gap (main ask)

  Go PR provides two options, Rust PR provides one.

* Go: `WebSocketConnectionHeaders(http.Header)`   Rust: `websocket_headers(http::HeaderMap)  `
* Go `WebSocketConnectionHeadersHandler(func() (http.Header, error))`    Rust: missing.

  Handler form invoked per-handshake in wsInitHandshake (ws.go:734). Real use case: refresh expiring
  bearer/JWT on reconnect. Static websocket_headers set once at builder time → tokens go stale, reconnects
  fail silently with 401 from proxy. Block parity.

  Suggested Rust shape (mirrors auth_callback pattern in options.rs):
  pub fn websocket_headers_callback<F, Fut>(mut self, cb: F) -> ConnectOptions
  where
      F: Fn() -> Fut + Send + Sync + 'static,
      Fut: Future<Output = Result<http::HeaderMap, AuthError>> + Send + 'static,
  Connector calls callback in the "ws" / "wss" arms instead of cloning a static map. Mirror Go's mutex
  check (ErrWebSocketHeadersAlreadySet) or make callback override static.
  
  
  Would be also really good to add some tests that check this path directly.
  
  Would be good to add a note that ws headers are single value and bump tokio websockets to 0.13

PR #1592, opened 2026-05-28
Add chrono as alternative to time.

vgaspary:
Both `time` and `chrono` are popular Rust crates providing datetime utilities. `async_nats` optionally depends on `time` for the "jetstream" feature.

This change adds `chrono` as another optional dependency, introduces the `datetime.rs` abstraction layer above `time` and `chrono`, allows the user to control which dependency to build `async_nats` with.

### DIFF
--- a/async-nats/src/datetime.rs
+++ b/async-nats/src/datetime.rs
@@ -1,0 +1,117 @@
+// Copyright 2020-2026
+// TODO
+
+#[cfg(all(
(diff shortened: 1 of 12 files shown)


async-nats/src/datetime.rs
```diff
@@ -0,0 +1,117 @@
+// Copyright 2020-2026
+// TODO
```
Jarema: please add the standard header

PR #1595, opened 2026-06-01
Add chrono as alternative to time crate

Jarema:
supersedes #1592

### DIFF
--- a/async-nats/src/datetime.rs
+++ b/async-nats/src/datetime.rs
@@ -1,0 +1,117 @@
+// Copyright 2020-2026
+// TODO
+
+#[cfg(all(
+    any(feature = "jetstream", feature = "service"),
+    not(feature = "time"),
+    not(feature = "chrono")
+))]
+compile_error!(
+    "Either 'time' or 'chrono' feature must be enabled when 'jetstream' or 'service' is active."
+);
+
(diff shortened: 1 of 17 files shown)


Jarema:
@claude review again

async-nats/src/datetime.rs
```diff
@@ -0,0 +1,329 @@
...
+//!
+//! - With `time` (default), [`DateTime`] is `time::OffsetDateTime`.
+//! - With `chrono` (opt-in), [`DateTime`] is `chrono::DateTime<chrono::Utc>`.
+//!
+//! When both features are enabled, `time` wins, so enabling `chrono`
+//! transitively never flips the public API type for crates on the default
+//! backend (Cargo features are additive). To build against `chrono`, disable
+//! default features and enable `chrono` explicitly.
+
+#[cfg(all(
```
caspervonb: I'd consider going with just one feature flag rather than a set of mutually exlusive, aiming to keep cargo test --all-features working.

If chromo feature is enabled, use chrono otherwise fallback to time. Use negatives in cfg.

caspervonb (commented):
One comment, not a blocker but worth considering.

caspervonb (approved):
Yeah feels better with a single feature gate, lgtm

PR #1599, opened 2026-06-04
Fix recreating ordered consumer on server restart

### LINKED ISSUE #1596: KV watch() errors with "missed idle heartbeat" and never resumes after a server restart, if the Watch stream wasn't being polled at restart   time (regressed 0.42 → 0.43)
# Summary

  After a nats-server restart, a JetStream KV Store::watch() does not resume if its Watch stream was not being polled at the moment of the restart.
  The first poll after the restart yields Err("watcher consumer error: missed idle heartbeat"), and every subsequent poll then hangs indefinitely —
  the ordered consumer is never recreated, post-restart updates are never delivered, and the stream never ends.

  A watch that is continuously polled (parked in .next().await) across the restart recovers correctly. The bucket is file-backed, so its stream/data
  survive the restart (entry() still returns values); only the watch fails to resume. This worked in 0.42.0 and regressed in 0.43.0.

  # Reproduction

  Deps: async-nats = "0.49", tokio = { version = "1", features = ["full"] }, futures = "0.3", bytes = "1".

  Start a server: nats-server -js -sd /tmp/nats-repro -p 4222

```
  use async_nats::jetstream::{self, kv};
  use bytes::Bytes;
  use futures::StreamExt;
  use std::time::Duration;

  #[tokio::main]
  async fn main() -> Result<(), Box<dyn std::error::Error>> {
      let js = jetstream::new(async_nats::connect("127.0.0.1:4222").await?);
      let bucket = js.create_key_value(kv::Config {
          bucket: "REPRO".into(), history: 5, ..Default::default()
      }).await?;

      // Open the watch and warm it with one live delivery (establishes the consumer).
      let mut watch = bucket.watch("k").await?;
      bucket.put("k", Bytes::from_static(b"v0")).await?;
      println!("warm: {:?}", tokio::time::timeout(Duration::from_secs(3), watch.next()).await?);
  
      // >>> Now SIGKILL nats-server and relaunch it on the SAME -sd dir. <<<
      // The watch is idle here — no .next() in flight.
      println!("kill + restart nats-server now…");
      tokio::time::sleep(Duration::from_secs(8)).await;

      // Publish after the restart, then resume polling for up to 15s.
      bucket.put("k", Bytes::from_static(b"v1")).await?;
      let deadline = tokio::time::Instant::now() + Duration::from_secs(15);
      loop {
          let rem = deadline.saturating_duration_since(tokio::time::Instant::now());
          if rem.is_zero() { println!("NEVER RECOVERED (v1 not delivered)"); break; }
          match tokio::time::timeout(rem, watch.next()).await {
              Ok(Some(Ok(e))) if e.value == "v1" => { println!("RECOVERED"); break; }
              Ok(Some(Ok(_)))  => continue,
              Ok(Some(Err(e))) => println!("stream error: {e}"),   // "missed idle heartbeat"
              Ok(None)         => { println!("stream ended"); break; }
              Err(_)           => { println!("NEVER RECOVERED (hung)"); break; }
          } 
      }
      Ok(())
  }
```
  
  ## Observed

```
  warm: Ok(Some(Ok(Entry { value: b"v0", .. })))
  stream error: watcher consumer error: missed idle heartbeat
  NEVER RECOVERED (hung)      # subsequent polls never yield; v1 lost; stream never ends
```

 ## Expected

  The watcher recreates its ordered consumer from the last delivered sequence and delivers v1 — as in 0.42.0. At minimum, the missed-heartbeat condition should trigger consumer recreation rather than a terminal error followed by an indefinite hang.

  ## Bisect
```
  ┌─────────────────┬───────────────────────────────
────────┐
  │   async-nats    │    idle-at-restart watch recovers?    │
  ├─────────────────┼───────────────────────────────────────┤
  │ 0.42.0          │ ✅                                    │
  ├─────────────────┼───────────────────────────────────────┤
  │ 0.43.0 – 0.49.0 │ ❌ (missed idle heartbeat, then hang) │
  └─────────────────┴───────────────────────────────────────┘
```

  ## Contrast (what does work)

  A watch kept continuously parked in .next().await across the restart recovers and delivers v1 — so recreation works while the stream is being 
  polled, but not when polling resumes after an idle period that spanned the restart.

  ## Suspected cause

  KV watch() uses an ephemeral ordered push consumer; consumer recreation is driven by the idle-heartbeat monitor, which (per #1046/#1048) only
  advances while the Stream is polled. So a watcher idle at restart never detects the loss live, and the first resumed poll surfaces a terminal
  missed idle heartbeat instead of recreating. (Cross-client: nats.go #1094, nats.py #510, nats.java #924, nats.net #202.)

 ## Client-side workaround (for others hitting this)
  
  Reopening the watch on the error with watch_with_history() (DeliverPolicy::LastPerSubject) + revision-dedup recovers the missed value. Reopening
  with plain watch() (DeliverPolicy::New) does not — it drops the update written during the outage.

 ## Questions

  1. Is the active-polling requirement for KV-watch recovery intended (a consequence of #1046/#1048), or a regression from 0.42?
  2. If intended, what's the recommended pattern for a KV watch that must survive restarts without being continuously polled — e.g. recreate the
  consumer on Event::Connected, or use a durable consumer on the KV-backing stream?

### Expected behavior

A watch kept continuously parked in .next().await across the restart recovers and delivers v1 — so recreation works while the stream is being polled, but not when polling resumes after an idle period that spanned the restart. Can polling after restart also recover without workaround?

### Server and client version

async-nats 0.49.0 (latest); also reproduced on 0.43.0–0.48.0. Last good: 0.42.0.
nats-server 2.14.0, file-backed JetStream (-js -sd ).
Rust 1.94, macOS/Linux.

### Host environment

_No response_

### Steps to reproduce

_No response_

Jarema:
While removing additional task in v0.43, we accidentaly introduced a regression taht in cases described in #1596 could cause a hanging push consumer.

fix #1596 

### DIFF
--- a/async-nats/src/jetstream/consumer/push.rs
+++ b/async-nats/src/jetstream/consumer/push.rs
@@ -549,25 +549,6 @@ impl futures_util::Stream for Ordered {
 
     fn poll_next(mut self: Pin<&mut Self>, cx: &mut task::Context<'_>) -> Poll<Option<Self::Item>> {
         loop {
-            // Check heartbeat timer, but only if we have an active subscriber
-            // Don't check during recreation
-            if self.subscriber.is_some()
-                && self.subscriber_future.is_none()
-                && self
-                    .heartbeat_sleep
-                    .get_or_insert_with(|| {
-                        Box::pin(tokio::time::sleep(ORDERED_IDLE_HEARTBEAT.saturating_mul(2)))
-                    })
-                    .poll_unpin(cx)
-                    .is_ready()
-            {
-                self.heartbeat_sleep = None;
-                self.subscriber = None;
-                self.consumer_sequence.store(0, Ordering::Relaxed);
-                return Poll::Ready(Some(Err(OrderedError::new(
-                    OrderedErrorKind::MissingHeartbeat,
-                ))));
-            }
             // Poll for connection state changes
             match &mut self.state_change_future {
                 None => {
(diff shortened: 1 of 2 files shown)


caspervonb (commented):
Valid edge case raised by Codex I think, drain before discard?

Jarema:
@caspervonb yes, moving the logic to Pending - when everything is drained.

caspervonb (approved):
lgtm

Jarema:
@claude review again

PR #1601, opened 2026-06-12
Replace dead feature-powerset CI job with curated feature check

Jarema:
## Problem

The `check_features` job in `main.yml` runs `cargo hack check --feature-powerset`. With ~18 features the powerset explodes into hundreds of thousands of combinations, and the job has been cancelled at the 6h GitHub Actions cap on **every** recent push to main — feature-combination checking has never actually passed, while burning 6h of CI per push. Meanwhile `async-nats/.cargo-hack-check.sh` (written precisely to avoid the explosion) was referenced by no workflow.

This is exactly how feature-gated breakage can ship undetected (e.g. cfg-gated builder code only compiled under specific flag sets).

## Changes

- `main.yml`: `check_features` now runs `bash .cargo-hack-check.sh` with `timeout-minutes: 45` instead of the powerset.
- `test.yml`: new `check_features` job — feature combinations are now checked on pull requests, not only after merge.
- `.cargo-hack-check.sh`: extended for recent feature additions — `server_2_14` (now in defaults), `fips`, and `experimental`. Chrono-backend combinations were already covered.

## Verification

Full script run locally with `RUSTFLAGS="-D warnings"`: all strategies pass (`--each-feature` with ring, curated combinations incl. chrono backend, aws-lc-rs/fips, all server version flags, experimental).

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -59,6 +59,7 @@ jobs:
         working-directory: ./async-nats
     name: check feature flags (ubuntu-latest / stable)
     runs-on: ubuntu-latest
+    timeout-minutes: 45
     steps:
       - name: Check out repository
         uses: actions/checkout@v3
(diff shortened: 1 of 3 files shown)


piotrpio (approved):
LGTM!

PR #1602, opened 2026-06-12
Bound Error's From conversion with a local ErrorKind marker trait

Jarema:
## Problem

`time` 0.3.48 added an internal `impl From<HourBase> for <HourBase as ModifierValue>::Type`. Downstream coherence cannot see through the projection self type, so rustc conservatively reports E0119 against our blanket `impl<Kind: Clone + Debug + Display + PartialEq> From<Kind> for Error<Kind>` (`src/error.rs`). Since `Cargo.lock` is not committed, every fresh CI resolve picks 0.3.48 and **all CI jobs fail to compile**. Upstream: https://github.com/time-rs/time/issues/783

The deeper issue is ours: a `From` impl generic over all types matching std bounds makes a universal coherence claim — *any* dependency can collide with it, not just `time`.

## Change

Replace the open generic bound with a local marker trait (no version pin):

```rust
pub trait ErrorKind: Clone + Debug + Display + PartialEq {}

impl<Kind: ErrorKind> From<Kind> for Error<Kind> { ... }
```

Upstream crates cannot implement a local trait for their types (orphan rule) and the compiler knows every local impl concretely, so overlap is provably impossible — the crate compiles against `time` 0.3.48 unpinned, and is immune to this class of breakage from any future dependency release.

A crate-internal `error_kinds!` macro implements the marker for all kind enums (one variadic invocation per module, cfg-split for the `kv`/`object-store`/`server_2_14`-gated kinds). The trait is public and unsealed with a doc example, so downstream users of `Error<CustomKind>` keep the `kind.into()` conversion by adding one impl line — that one-liner is the only (pre-1.0-acceptable) breaking edge.

## Verification

All run with `time` resolved to the breaking 0.3.48:
- full `.cargo-hack-check.sh` feature sweep (30 builds) ✅
- `cargo clippy --benches --tests --examples --all-features -- --deny clippy::all` ✅
- unit tests: 86 passed ✅

### DIFF
--- a/async-nats/CHANGELOG.md
+++ b/async-nats/CHANGELOG.md
@@ -1,4 +1,15 @@
 # Unreleased
+## Breaking changes
+* The generic `From<Kind> for Error<Kind>` conversion is now bounded by the new
+  `async_nats::error::ErrorKind` marker trait instead of plain
+  `Clone + Debug + Display + PartialEq`. The unbounded conversion conflicted (E0119) with
+  `time` 0.3.48 on Rust 1.84+ and any future dependency shipping a projection-self-type
+  `From` implementation could do the same. All built-in kind enums implement the marker, so
+  code using async-nats' own error types is unaffected. If you define custom kinds with
+  `async_nats::error::Error<MyKind>` and convert via `kind.into()`, add:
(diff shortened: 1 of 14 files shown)


caspervonb:
Okay so we are trying to fix a symptom here, makes me wonder if we shouldn't just drop the from implementation and instead be explicit?

PR #1604, opened 2026-06-12
fix account info deser failure on servers with tiered jetstream

xanderio:
The `api` field doesn't exist in the tier response object, this was confirmed by looking at the relavent go code [1].

nats-servers without tiered jetstream, it seems that this is only included if JWT based authentiction is in use, don't include the `tier` object and field at all. That allows clients to connect without running into the deser error.

[1]: https://github.com/nats-io/nats-server/blob/717c7d17efb387b6db31ed3a0301d1250b073207/server/jetstream.go#L77

### DIFF
--- a/async-nats/src/jetstream/account.rs
+++ b/async-nats/src/jetstream/account.rs
@@ -85,9 +85,6 @@ pub struct Tier {
     pub consumers: usize,
     /// Limits imposed on this tier.
     pub limits: Limits,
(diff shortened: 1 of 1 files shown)


Jarema:
Thanks for the catch and fix. Looks correct. Witing for CI.

Jarema (approved):
LGTM!

Thanks for the contribution!

PR #1606, opened 2026-06-16
Populate ObjectInfo modified from server message timestamp

### LINKED ISSUE #1605: ObjectStore: modified (mtime) read from JSON payload instead of server message timestamp
### Observed behavior

`ObjectInfo.modified` (mtime) returned by the object store is populated from the
`mtime` field deserialized out of the meta message JSON payload, instead of from
the server-assigned message timestamp. Both read paths do this:

- `ObjectStore::info()` (`async-nats/src/jetstream/object_store/mod.rs:258`)
  deserializes `ObjectInfo` from `message.payload` and returns it untouched. The
  `StreamMessage` it fetched carries `message.time` (server publish time), which
  is discarded.
- `Lister::poll_next`, used by `list()` (same file, ~L899–912), already extracts
  `message.info()` and reads `info.pending`, but deserializes `modified` from the
  payload and ignores `info.published` (the server receive time).

Because `modified` is `Option<DateTime>` with `#[serde(default)]`, an object
written by a client that stores a zero or absent `mtime` deserializes in Rust to
either `None` or `Some(0001-01-01T00:00:00Z)`, depending on which client wrote
the object.

This surfaced while investigating the Go side in nats-io/nats.go#2096, where the
maintainer confirmed the fix belongs in the Rust client:

> ObjectInfo.ModTime returned to the user should not be populated from the json,
> but rather from msg timestamp metadata — see ADR-20. [...] the fix should be
> applied in rust client so that it properly populates ModTime with msg timestamp.

### Expected behavior

Per [ADR-20 (Object Store)](https://github.com/nats-io/nats-architecture-and-design/blob/main/adr/ADR-20.md),
the `modified` (ModTime) value returned to the user should come from the message
timestamp, not from the stored JSON `mtime`. Other clients (e.g. nats.go) already
substitute the server timestamp on read; `async-nats` should do the same on both
read paths, so `info()` and `list()` agree and return a non-zero timestamp
regardless of which client wrote the object.

### Server and client version

- async-nats: current `main` (reproduces on the latest published release as well;
  the affected code is unchanged on `main`)
- nats-server: not version-specific — the timestamp is present on the JetStream
  message metadata for all supported server versions (2.10+)

### Host environment

N//A this is a logic bug so it is host independent

### Steps to reproduce

With a running `nats-server` (JetStream enabled), have the Go client write an
object and the Rust client read it back.

**Go (nats.go v1.52.0) writes the object:**
```go
js, _ := jetstream.New(nc)
os, _ := js.CreateObjectStore(ctx, jetstream.ObjectStoreConfig{Bucket: "xbucket"})
os.Put(ctx, jetstream.ObjectMeta{Name: "FOO"}, bytes.NewReader([]byte{2, 3, 4, 5}))
// stored meta JSON contains: "mtime":"0001-01-01T00:00:00Z"
```

**Rust (async-nats) reads the same object:**
```rust
let bucket = js.get_object_store("xbucket").await?;
println!("{:?}", bucket.info("FOO").await?.modified);
```

**Result:**
```
RUST info().modified: Some(0001-01-01 00:00:00 +00:00)
RUST list().modified: Some(0001-01-01 00:00:00 +00:00)
```

Prashant-koi:
Fixes #1605 

## Overview

`ObjectStore::info()` and `list()` populate `ObjectInfo.modified` directly from
the `mtime` field of the meta message JSON payload. Per [ADR-20](https://github.com/nats-io/nats-architecture-and-design/blob/main/adr/ADR-20.md#objectinfo-storage), the value
returned to the user should be the server-assigned message timestamp, not the
stored `mtime`, which some clients (e.g. nats.go) leave zeroed in the JSON.

As a result, an object written by such a client is read back in async-nats with
`modified` set to the zero time (`0001-01-01T00:00:00Z`) instead of the real
modification time, breaking cross-client parity.

## Changes

- `info()`: substitute `message.time` (the server publish time, already fetched)
  into `modified` after deserializing.
- `Lister::poll_next`: substitute `info.published` into `modified` (copied out
  before the payload borrow, since `info` borrows the message).
- Add `modified_cross_client_parity` test: simulates a foreign writer storing a
  zeroed `mtime` and asserts both read paths return the server timestamp.

Both paths use timestamps that were already being fetched and discarded, so
there should be no extra round-trip.

Signed-off-by: Prasant Koirala <koiralaprashanta10@gmail.com>

### DIFF
--- a/async-nats/src/jetstream/object_store/mod.rs
+++ b/async-nats/src/jetstream/object_store/mod.rs
@@ -255,13 +255,16 @@ impl ObjectStore {
                 }
                 _ => InfoError::with_source(InfoErrorKind::Other, err),
             })?;
-        let object_info =
+        let mut object_info =
             serde_json::from_slice::<ObjectInfo>(&message.payload).map_err(|err| {
                 InfoError::with_source(
                     InfoErrorKind::Other,
                     format!("failed to decode info payload: {err}"),
                 )
             })?;
+        // Per ADR-20, `modified` is the server message timestamp, not the `mtime`
+        // stored in the JSON payload (which some clients leave zeroed)
+        object_info.modified = Some(message.time);
 
         Ok(object_info)
     }
@@ -903,13 +906,19 @@ impl Stream for List {
                             if info.pending == 0 {
                                 self.done = true;
                             }
-                            let response: ObjectInfo = serde_json::from_slice(&message.payload)
+                            // Copy the server timestamp out before borrowing the
+                            // payload below, since `info` borrows `message`
+                            let modified = info.published;
+                            let mut response: ObjectInfo = serde_json::from_slice(&message.payload)
                                 .map_err(|err| {
                                     ListerError::with_source(
                                         ListerErrorKind::Other,
                                         format!("failed deserializing object info: {err}"),
                                     )
                                 })?;
+                            // Per ADR-20, `modified` is the server message timestamp,
+                            // not the `mtime` stored in the JSON payload
+                            response.modified = Some(modified);
                             if response.deleted {
                                 continue;
                             }
--- a/async-nats/tests/object_store.rs
+++ b/async-nats/tests/object_store.rs
@@ -319,6 +319,81 @@ mod object_store {
     }
 
     #[tokio::test]
+    async fn modified_cross_client_parity() {
+        // Guards the cross-client parity fix: a foreign client (e.g. nats.go) may
+        // store a meta message whose JSON `mtime` is the zero time. Both Rust read
+        // paths must return the server message timestamp instead of that zeroed
+        // `mtime`, matching ADR-20 and the behavior of other clients
+        let server = nats_server::run_server("tests/configs/jetstream.conf");
+        let client = async_nats::connect(server.client_url()).await.unwrap();
+        let jetstream = async_nats::jetstream::new(client);
+
+        let bucket = jetstream
+            .create_object_store(async_nats::jetstream::object_store::Config {
+                bucket: "bucket".to_string(),
+                ..Default::default()
+            })
+            .await
+            .unwrap();
+
+        // Simulate a foreign writer, this will publish a meta message with mtime = zero time
+        let name = "FOO";
+        let encoded = base64::engine::general_purpose::URL_SAFE.encode(name);
+        let subject = format!("$O.bucket.M.{encoded}");
+        let payload = format!(
+            r#"{{"name":"{name}","bucket":"bucket","nuid":"zzzzzzzzzzzzzzzzzzzzzz","size":4,"chunks":1,"mtime":"0001-01-01T00:00:00Z"}}"#
+        );
+        let mut headers = HeaderMap::new();
+        headers.insert("Nats-Rollup", "sub");
+        jetstream
+            .publish_with_headers(subject.clone(), headers, payload.clone().into())
+            .await
+            .unwrap()
+            .await
+            .unwrap();
+
+        // The zero time as parsed from that JSON, in the crate's DateTime type
+        // (chrono or time depending on features). This is what a buggy read would return.
+        let zero = serde_json::from_str::<ObjectInfo>(&payload)
+            .unwrap()
+            .modified
+            .unwrap();
+
+        // The server timestamp IS available on the stored message
+        let raw = jetstream
+            .get_stream("OBJ_bucket")
+            .await
+            .unwrap()
+            .get_last_raw_message_by_subject(&subject)
+            .await
+            .unwrap();
+        let server_time = raw.time;
+        println!("server message time: {server_time}");
+
+        // info() path
+        let info_modified = bucket.info(name).await.unwrap().modified;
+        println!("info().modified:     {info_modified:?}");
+
+        // list() path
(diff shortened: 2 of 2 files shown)


stasimus:
`async-nats/src/jetstream/object_store/mod.rs:861` Watch::poll_next is a third read path that deserializes ObjectInfo from the meta payload, and it does not get the same override. So watch() still returns the raw JSON mtime, and for a foreign writer with a zeroed mtime (the exact scenario in the new test) it will disagree with what info() and list() report for the same object. The fix applies there too: the push::Ordered message exposes message.info().published, same as List uses.

stasimus:
`async-nats/src/jetstream/object_store/mod.rs:189` Side effect worth flagging: delete() re-serializes the ObjectInfo it gets from the now-patched info() and publishes it as the rollup tombstone. Since modified serializes back out as mtime, the previous meta message's server receive time now ends up written into the tombstone JSON, which no other client does. Anything that trusts the payload mtime (watch in this crate, other language clients, older versions of this one) sees a fabricated modification time on the delete record.

stasimus:
`async-nats/src/jetstream/object_store/mod.rs:630` update_metadata() has the same write-side leak, plus it now returns a stale value. The ObjectInfo it serializes and returns carries the OLD meta message's server time, while info() or list() called right after report the NEW message's time. Before this change the two agreed. Code that caches the return value for change detection will see a phantom modification on the next read.

Jarema:
Taking a look

stasimus:
`async-nats/src/jetstream/object_store/mod.rs:366` put() (and add_link / add_bucket_link) still returns modified = client-local now(), which after this change systematically disagrees with the server timestamp that info() and list() return for the same object version. With a client clock a few seconds behind, put()'s timestamp can even sort after a later overwrite's server time. Before the PR the two matched exactly, since info() read back the mtime that put() wrote.

async-nats/src/jetstream/object_store/mod.rs
```diff
@@ -255,13 +255,16 @@ impl ObjectStore {
...
+        let mut object_info =
             serde_json::from_slice::<ObjectInfo>(&message.payload).map_err(|err| {
                 InfoError::with_source(
                     InfoErrorKind::Other,
                     format!("failed to decode info payload: {err}"),
                 )
             })?;
+        // Per ADR-20, `modified` is the server message timestamp, not the `mtime`
+        // stored in the JSON payload (which some clients leave zeroed)
+        object_info.modified = Some(message.time);
```
stasimus, line 267: This override is applied per call site (here and again in List), which is how the watch path got missed and how the write paths end up leaking the overridden value back into the JSON. A single helper like object_info_from_message(payload, server_time), or serde-level handling of the mtime field, would fix the invariant once. The FIXME above about migrating info() to a direct get is a concrete future edit that would otherwise have to remember to re-apply this by hand.

async-nats/src/jetstream/object_store/mod.rs
```diff
@@ -903,13 +906,19 @@ impl Stream for List {
                             if info.pending == 0 {
                                 self.done = true;
                             }
-                            let response: ObjectInfo = serde_json::from_slice(&message.payload)
+                            // Copy the server timestamp out before borrowing the
+                            // payload below, since `info` borrows `message`
+                            let modified = info.published;
```
stasimus, line 911: The comment is not actually true: info() takes &self and from_slice(&message.payload) is also a shared borrow, so `response.modified = Some(info.published)` compiles fine without the pre-copy (Info.published is a Copy DateTime). Either drop the intermediate variable or fix the comment, otherwise it teaches a borrow rule that does not exist.

async-nats/tests/object_store.rs
```diff
@@ -318,6 +318,81 @@ mod object_store {
...
+        // The server timestamp IS available on the stored message
+        let raw = jetstream
+            .get_stream("OBJ_bucket")
+            .await
+            .unwrap()
+            .get_last_raw_message_by_subject(&subject)
+            .await
+            .unwrap();
+        let server_time = raw.time;
+        println!("server message time: {server_time}");
```
stasimus, line 371: Leftover debug printlns (here and lines 375, 380)? The asserts below already print both values with descriptive messages when they fail, and the rest of this file has basically none.

async-nats/tests/object_store.rs
```diff
@@ -318,6 +318,81 @@ mod object_store {
...
+
+        // Simulate a foreign writer, this will publish a meta message with mtime = zero time
+        let name = "FOO";
+        let encoded = base64::engine::general_purpose::URL_SAFE.encode(name);
+        let subject = format!("$O.bucket.M.{encoded}");
+        let payload = format!(
+            r#"{{"name":"{name}","bucket":"bucket","nuid":"zzzzzzzzzzzzzzzzzzzzzz","size":4,"chunks":1,"mtime":"0001-01-01T00:00:00Z"}}"#
+        );
+        let mut headers = HeaderMap::new();
+        headers.insert("Nats-Rollup", "sub");
```
stasimus, line 347: There is a public constant for this: async_nats::header::NATS_ROLLUP. Worth using it here, since a typo in the literal would turn this into a non-rollup publish and the test would still pass for the wrong reason (only one meta message exists on the subject anyway).

PR #1608, opened 2026-07-01
Avoid panic in Subscriber::drop when dropped outside a runtime

### LINKED ISSUE #1607: async-nats: Subscriber::drop panics when dropped outside a Tokio runtime
## What

`Subscriber::drop` unconditionally calls `tokio::spawn` to send the final `Unsubscribe` command:

https://github.com/nats-io/nats.rs/blob/main/async-nats/src/lib.rs#L1421-L1435

```rust
impl Drop for Subscriber {
    fn drop(&mut self) {
        self.receiver.close();
        tokio::spawn({
            let sender = self.sender.clone();
            let sid = self.sid;
            async move {
                sender
                    .send(Command::Unsubscribe { sid, max: None })
                    .await
                    .ok();
            }
        });
    }
}
```

`tokio::spawn` **panics** when called outside a runtime context (`there is no reactor running, must be called from the context of a Tokio 1.x runtime`). `Drop` can run on arbitrary threads, during unwinding, and at teardown — situations where no runtime is entered — so a `Subscriber` dropped in any of those contexts aborts.

## Reproduction

```rust
#[test]
fn drop_subscriber_outside_runtime() {
    let server = nats_server::run_basic_server();
    let rt = tokio::runtime::Runtime::new().unwrap();
    let subscription = rt.block_on(async {
        let client = async_nats::connect(server.client_url()).await.unwrap();
        client.subscribe("foo").await.unwrap()
    });
    // Outside the runtime context now — this drop panics.
    drop(subscription);
}
```

Any code holding a `Subscriber` in a struct that is later dropped from a non-async context (e.g. a synchronous `main` after the runtime handle is out of scope, a `std::thread`, or during panic unwinding) hits this.

## Affected versions

Present on `main` and in the latest release (`async-nats` v0.49.1). Dates back to the original 2022 implementation.

## Suggested fix

Drop-based cleanup should be synchronous and best-effort. The crate already does this elsewhere — `PublishAckFuture::drop` uses `try_send`, and the connection handler fires events via `try_send(...).ok()`. Replacing the spawn with a `try_send` removes the runtime dependency entirely (no `Handle::try_current()` guard needed) and is consistent with those idioms. Note that a `tokio::spawn`-based cleanup is also unreliable at teardown regardless of the panic: a detached task on a shutting-down runtime may never be polled to completion, so `send().await` provides no real delivery guarantee over `try_send` there.

The reliable, back-pressured path already exists for callers who need it: `Subscriber::unsubscribe().await` / `drain().await`.

I have a fix + regression test ready and will open a PR.

lives-hoth:
Fixes #1607.

## Problem

`Subscriber::drop` unconditionally called `tokio::spawn` to send the final `Unsubscribe` command:

```rust
tokio::spawn({
    let sender = self.sender.clone();
    let sid = self.sid;
    async move { sender.send(Command::Unsubscribe { sid, max: None }).await.ok(); }
});
```

`tokio::spawn` panics when called outside a Tokio runtime context (`there is no reactor running, must be called from the context of a Tokio 1.x runtime`). `Drop` can run on arbitrary threads, during unwinding, and at teardown — so a `Subscriber` dropped in any of those contexts aborts.

## Fix

Replace the spawn with a synchronous, runtime-independent `try_send`, which removes the runtime dependency entirely (no `Handle::try_current()` guard needed). This matches the crate's existing best-effort `Drop` idioms (`PublishAckFuture::drop`, the connection handler's event sends).

On a full or closed command channel the unsubscribe is dropped and logged at `debug`; the server retains interest only until the next unsubscribe or disconnect, so nothing lasting leaks. `debug` (not `warn`) mirrors how the message-send path handles the same `TrySendError` cases: a closed channel is routine teardown, and channel saturation is already surfaced via `Event::SlowConsumer`. Callers needing guaranteed, back-pressured delivery still have `Subscriber::unsubscribe().await` / `drain().await`.

## Test

Adds `drop_subscriber_outside_runtime`, which builds the client inside a runtime and drops the subscriber outside the runtime context. It panics on the old implementation and passes with the fix.

### DIFF
--- a/async-nats/src/lib.rs
+++ b/async-nats/src/lib.rs
@@ -1421,16 +1421,18 @@ impl From<tokio::sync::mpsc::error::SendError<Command>> for UnsubscribeError {
 impl Drop for Subscriber {
     fn drop(&mut self) {
         self.receiver.close();
-        tokio::spawn({
-            let sender = self.sender.clone();
-            let sid = self.sid;
-            async move {
-                sender
-                    .send(Command::Unsubscribe { sid, max: None })
-                    .await
-                    .ok();
-            }
-        });
+        // Best-effort unsubscribe. `tokio::spawn` would panic here since `Drop` can run outside
+        // any runtime (unwinding, teardown, a plain thread), so use synchronous `try_send`.
+        // If the channel is full or closed the unsubscribe is dropped; the server keeps interest
+        // only until the next unsubscribe or disconnect, so nothing lasting leaks.
+        // Debug, not warn: a closed channel is routine teardown and saturation already surfaces as
+        // `Event::SlowConsumer`.
+        if let Err(err) = self.sender.try_send(Command::Unsubscribe {
+            sid: self.sid,
+            max: None,
+        }) {
+            debug!("failed to send unsubscribe in Subscriber::drop: {err}");
+        }
     }
 }
 
--- a/async-nats/src/lib.rs
+++ b/async-nats/src/lib.rs
@@ -1437,18 +1437,27 @@ impl Drop for Subscriber {
     fn drop(&mut self) {
         self.receiver.close();
         // Best-effort unsubscribe. `Drop` can run outside a Tokio runtime (unwinding, teardown,
-        // plain threads), so only spawn when a runtime is available and otherwise use
-        // synchronous `try_send` to avoid panics.
-        if let Ok(handle) = tokio::runtime::Handle::try_current() {
-            let sender = self.sender.clone();
-            let sid = self.sid;
-            let _ = handle.spawn(async move {
-                if let Err(err) = sender.send(drop_unsubscribe_command(sid)).await {
-                    debug!("failed to send unsubscribe in Subscriber::drop: {err}");
+        // plain threads), so start with synchronous `try_send` and only spawn an async retry when
+        // the command channel is full and a runtime is available.
+        match self.sender.try_send(drop_unsubscribe_command(self.sid)) {
+            Ok(()) => {}
+            Err(tokio::sync::mpsc::error::TrySendError::Full(command)) => {
+                if let Ok(handle) = tokio::runtime::Handle::try_current() {
+                    let sender = self.sender.clone();
+                    std::mem::drop(handle.spawn(async move {
+                        if let Err(err) = sender.send(command).await {
+                            debug!("failed to send unsubscribe in Subscriber::drop: {err}");
+                        }
+                    }));
+                } else {
+                    debug!(
+                        "failed to send unsubscribe in Subscriber::drop: command channel full and no runtime"
+                    );
                 }
-            });
-        } else if let Err(err) = self.sender.try_send(drop_unsubscribe_command(self.sid)) {
-            debug!("failed to send unsubscribe in Subscriber::drop: {err}");
+            }
+            Err(tokio::sync::mpsc::error::TrySendError::Closed(_)) => {
+                debug!("failed to send unsubscribe in Subscriber::drop: command channel closed");
+            }
         }
     }
 }
--- a/async-nats/src/lib.rs
+++ b/async-nats/src/lib.rs
@@ -1275,13 +1275,20 @@ impl From<io::Error> for ConnectError {
 /// Retrieves messages from given `subscription` created by [Client::subscribe].
 ///
 /// Implements [futures_util::stream::Stream] for ergonomic async message processing.
+/// For deterministic cleanup, prefer calling [`Subscriber::unsubscribe`] or
+/// [`Subscriber::drain`]. Dropping a [`Subscriber`] is best-effort.
 ///
 /// # Examples
-/// ```
+/// ```no_run
+/// # use futures_util::StreamExt;
 /// # #[tokio::main]
 /// # async fn main() ->  Result<(), async_nats::Error> {
-/// let mut nc = async_nats::connect("demo.nats.io").await?;
-/// # nc.publish("test", "data".into()).await?;
+/// let client = async_nats::connect("demo.nats.io").await?;
+/// let mut subscriber = client.subscribe("events.>").await?;
+///
+/// while let Some(message) = subscriber.next().await {
+///     println!("received: {message:?}");
+/// }
 /// # Ok(())
 /// # }
 /// ```
@@ -1308,7 +1315,7 @@ impl Subscriber {
     /// Unsubscribes from subscription, draining all remaining messages.
     ///
     /// # Examples
-    /// ```
+    /// ```no_run
     /// # #[tokio::main]
     /// # async fn main() -> Result<(), async_nats::Error> {
     /// let client = async_nats::connect("demo.nats.io").await?;
@@ -1319,6 +1326,9 @@ impl Subscriber {
     /// # Ok(())
     /// # }
     /// ```
+    ///
+    /// This method is the reliable unsubscribe path: it waits for the command to be accepted by
+    /// the client's command channel. Dropping [`Subscriber`] is best-effort.
     pub async fn unsubscribe(&mut self) -> Result<(), UnsubscribeError> {
         self.sender
             .send(Command::Unsubscribe {
@@ -1331,11 +1341,11 @@ impl Subscriber {
     }
 
     /// Unsubscribes from subscription after reaching given number of messages.
-    /// This is the total number of messages received by this subscription in it's whole
+    /// This is the total number of messages received by this subscription in its whole
     /// lifespan. If it already reached or surpassed the passed value, it will immediately stop.
     ///
     /// # Examples
-    /// ```
+    /// ```no_run
     /// # use futures_util::StreamExt;
     /// # #[tokio::main]
     /// # async fn main() -> Result<(), async_nats::Error> {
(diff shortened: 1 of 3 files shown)


async-nats/src/lib.rs
```diff
@@ -1421,16 +1421,18 @@ impl From<tokio::sync::mpsc::error::SendError<Command>> for UnsubscribeError {
...
+        // Best-effort unsubscribe. `tokio::spawn` would panic here since `Drop` can run outside
+        // any runtime (unwinding, teardown, a plain thread), so use synchronous `try_send`.
+        // If the channel is full or closed the unsubscribe is dropped; the server keeps interest
+        // only until the next unsubscribe or disconnect, so nothing lasting leaks.
+        // Debug, not warn: a closed channel is routine teardown and saturation already surfaces as
+        // `Event::SlowConsumer`.
+        if let Err(err) = self.sender.try_send(Command::Unsubscribe {
+            sid: self.sid,
+            max: None,
+        }) {
```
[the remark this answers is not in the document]
lives-hoth: Thanks — this is a real tradeoff, and it was a deliberate choice rather than an oversight.

- **`Full` is rare by default.** `sender_capacity` defaults to 2048, so `try_send` only returns `Full` when the client's command channel is already saturated — i.e. the connection is back-pressured or wedged. Spawning a task in that state doesn't make the UNSUB more likely to reach the server; it just defers the same congestion.
- **The spawn path isn't reliable where `Drop` actually runs.** `Subscriber` is most often dropped at teardown, where the runtime may be shutting down. A detached `tokio::spawn` on a stopping runtime may never be polled to completion, so `send().await` provides no real delivery guarantee in exactly the case that motivated removing the spawn (it also panics off-runtime, which is the bug this PR fixes).
- **The leak is bounded and self-healing.** `receiver.close()` runs first, so a dropped UNSUB leaks only *server-side interest*, and only until the next unsubscribe on that sid or the next disconnect. Messages are discarded client-side in the meantime; there's no client-side correctness impact.
- **Consistency + an explicit reliable path.** `try_send` matches the crate's existing best-effort `Drop` idioms (`PublishAckFuture::drop`, and the message-send path in `lib.rs` which handles the same `Full`/`Closed` split at debug). Callers who need guaranteed, back-pressured delivery already have `Subscriber::unsubscribe().await` / `drain().await`.

The hybrid (`Handle::try_current()` → spawn `send().await`, else `try_send`) would preserve back-pressured delivery in the in-runtime case, but at the cost of reintroducing a detached task with no completion guarantee at teardown and a second code path — so I've kept the single synchronous `try_send`. Happy to revisit if a maintainer would prefer the hybrid.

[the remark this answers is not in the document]
Jarema: Hey @lives-hoth .

Thanks for the PR!

> Thanks — this is a real tradeoff, and it was a deliberate choice rather than an oversight.
> 
> Full is rare by default. sender_capacity defaults to 2048, so try_send only returns Full when the client's command channel is already saturated — i.e. the connection is back-pressured or wedged. Spawning a task in that state doesn't make the UNSUB more likely to reach the server; it just defers the same congestion.

The NATS client needs to be resilient and work properly in all known edge cases.
In heavily congested setups `Full` is actually a common scenario, and zombie subscriptions in big topologies are a very problematic thing.

I agree that there is a problem here. Though I don't think this is a right solution.

> The spawn path isn't reliable where Drop actually runs. Subscriber is most often dropped at teardown, where the runtime may be shutting down. A detached tokio::spawn on a stopping runtime may never be polled to completion, so send().await provides no real delivery guarantee in exactly the case that motivated removing the spawn (it also panics off-runtime, which is the bug this PR fixes).

Unsubscribing is thing that happens often durign application lifetime. Not just during teardown. And reliability of teardown is a less issue, as server will drop the subscription for dropped connections. 

> The leak is bounded and self-healing. receiver.close() runs first, so a dropped UNSUB leaks only server-side interest, and only until the next unsubscribe on that sid or the next disconnect. Messages are discarded client-side in the meantime; there's no client-side correctness impact.

I'm not sure if I understand this point. Server side leaks is the biggest problem in this context.
Disconnect case is not that big of a deal. The big deal is long-living client with zombie subscription.
Those will never clean up.


> Consistency + an explicit reliable path. try_send matches the crate's existing best-effort Drop idioms (PublishAckFuture::drop, and the message-send path in lib.rs which handles the same Full/Closed split at debug). Callers who need guaranteed, back-pressured delivery already have Subscriber::unsubscribe().await / drain().await.

The difference come from a fact, that we can drop client-side effects usually just fine. Here, its about sending important command to the server.


**What you brought is right - it would be good to fix this issue, and the root cause is lack of `async drop`, so we agreed to go with this until `async drop` is available in Rust. However, I think we need to think about bit  different solution**

lives-hoth:
Thanks @Jarema — I really appreciate the review and the way you framed the reliability concerns here. I learned a lot from your comments, especially around long-lived congested topologies and why server-side zombie subscriptions are the risk to optimize for.

I pushed an update that tries to follow that guidance directly:

- `Subscriber::drop` now uses a hybrid path:
  - if a Tokio runtime is available, it uses async `send(...).await` (back-pressured)
  - if no runtime is available, it falls back to synchronous `try_send` (panic-free)
- this keeps the off-runtime panic fix while preserving the more reliable in-runtime unsubscribe behavior.

I also added targeted tests to cover the transition boundaries we discussed:

- `drop_subscriber_in_runtime_full_command_channel_eventually_unsubscribes`
- `drop_subscriber_outside_runtime_open_command_channel_sends_unsubscribe`
- `drop_subscriber_outside_runtime_full_command_channel_is_best_effort`
- `drop_subscriber_outside_runtime_closed_command_channel_does_not_panic`
- `drop_subscriber_in_runtime_closed_command_channel_does_not_panic`
- `dropped_subscriber_not_replayed_after_reconnect`

Plus the existing integration regression remains:

- `drop_subscriber_outside_runtime`

I also cleaned up docs to make the user contract clearer: explicit `unsubscribe().await` / `drain().await` are the deterministic path; `Drop` is still best-effort.

There is still one inherent limitation here (until async drop exists): if runtime teardown races the spawned task, delivery is still best-effort in that shutdown window. I wanted to call that out explicitly so we’re aligned on the remaining edge case.

Thanks again for the thoughtful review — this was very helpful.

async-nats/src/lib.rs
```diff
@@ -1418,19 +1428,28 @@ impl From<tokio::sync::mpsc::error::SendError<Command>> for UnsubscribeError {
...
+}
+
 impl Drop for Subscriber {
     fn drop(&mut self) {
         self.receiver.close();
-        tokio::spawn({
+        // Best-effort unsubscribe. `Drop` can run outside a Tokio runtime (unwinding, teardown,
+        // plain threads), so only spawn when a runtime is available and otherwise use
+        // synchronous `try_send` to avoid panics.
+        if let Ok(handle) = tokio::runtime::Handle::try_current() {
```
Jarema: We could improve it and start with `try_send`, and only if we got full channel, switch to spawning tasks. That would heavily limit number of tasks, as channel is rarely full.

async-nats/src/lib.rs
```diff
@@ -1418,19 +1428,28 @@ impl From<tokio::sync::mpsc::error::SendError<Command>> for UnsubscribeError {
...
             let sender = self.sender.clone();
             let sid = self.sid;
-            async move {
-                sender
-                    .send(Command::Unsubscribe { sid, max: None })
-                    .await
-                    .ok();
-            }
-        });
+            let _ = handle.spawn(async move {
```
Jarema: the `let _ = ` fails clippy. just drop the let.

Jarema (changes requested):
Sorry for late answer.

This is much better.
Two nits to address and we can merge.

lives-hoth:
Thanks again @Jarema — great nits, and thank you for the patience here.

I pushed a follow-up commit to address both points:

1. `Subscriber::drop` now does `try_send` first, and only falls back to spawn when the channel is `Full` and a runtime is available.
2. Replaced `let _ = handle.spawn(...)` with explicit detach via `std::mem::drop(handle.spawn(...))` to keep clippy happy.

I reran the targeted tests around these transitions after the change:

- `drop_subscriber_in_runtime_full_command_channel_eventually_unsubscribes`
- `drop_subscriber_outside_runtime_open_command_channel_sends_unsubscribe`
- `drop_subscriber_outside_runtime_full_command_channel_is_best_effort`
- `drop_subscriber_outside_runtime_closed_command_channel_does_not_panic`
- `drop_subscriber_in_runtime_closed_command_channel_does_not_panic`
- `dropped_subscriber_not_replayed_after_reconnect`
- `drop_subscriber_outside_runtime`

Really appreciate the guidance on balancing reliability and runtime-safety — I learned a lot from this review cycle.

async-nats/src/lib.rs
```diff
@@ -1418,19 +1428,37 @@ impl From<tokio::sync::mpsc::error::SendError<Command>> for UnsubscribeError {
...
-                    .ok();
+        // Best-effort unsubscribe. `Drop` can run outside a Tokio runtime (unwinding, teardown,
+        // plain threads), so start with synchronous `try_send` and only spawn an async retry when
+        // the command channel is full and a runtime is available.
+        match self.sender.try_send(drop_unsubscribe_command(self.sid)) {
+            Ok(()) => {}
+            Err(tokio::sync::mpsc::error::TrySendError::Full(command)) => {
+                if let Ok(handle) = tokio::runtime::Handle::try_current() {
+                    let sender = self.sender.clone();
+                    std::mem::drop(handle.spawn(async move {
```
Jarema: by dropping `let` I meant literally dropping the `let =`. 

I would do this here:

``` 
 handle.spawn(async move {
      if let Err(err) = sender.send(command).await {
          debug!("failed to send unsubscribe in Subscriber::drop: {err}");
      }
  });
```

Jarema (changes requested):
one last nit before ready for merge :)

lives-hoth:
Addressed the last nit exactly as requested: removed the explicit detach wrapper and now call `handle.spawn(async move { ... });` directly (no `let _ =`, no `std::mem::drop(...)`).

I also re-ran focused tests with `RUSTFLAGS=-D warnings` locally for this path (`drop_subscriber_outside_runtime`, `dropped_subscriber_not_replayed_after_reconnect`) and they pass.

Separately, I audited the latest failing CI run while working this nit: the failing jobs are in broad `test` matrices with multiple JetStream failures (`too few tokens`), which I could not reproduce locally on current head (`jetstream_tests` passes locally).

async-nats/src/lib.rs
lives-hoth: Thanks for the clarification @Jarema — and thank you again for the careful review.

I took your suggestion literally and updated this to `handle.spawn(async move { ... });` (removed both `let _ =` and the explicit `drop(...)` wrapper).

I also re-ran the focused drop-path tests after that change and they pass locally. I really appreciate the guidance throughout this thread — I learned a lot from your feedback on keeping this path both runtime-safe and idiomatic.

PR #1610, opened 2026-07-21
Add docs.nats.io examples to main

Jarema:
Moves the documentation examples from the `core-docs` and `jetstream-docs` branches into `async-nats/examples/` on main (52 new `docs_learn_*.rs` cargo examples joining the 18 `docs_*` already here; additive only).

Why: the new docs.nats.io build fetches these snippets at build time. On side branches they get zero CI; on main they're compiled by `check_examples`/clippy/`build --all-targets` automatically. Verified locally: `cargo check --examples -p async-nats` passes, `cargo +nightly fmt` applied.

Note: please keep the docs branches until the docs repo's fetch config is switched to `main` (follow-up PR there).


piotrpio (approved):
LGTM

Issue #1548, opened 2026-03-20
Inconsistent construction of `HeaderValue`


gammelalf:
### Observed behavior

There is an `From<&str>` and a `FromStr` impl for `HeaderValue`.

The first would suggest the conversion to be infallible while the latter does the opposite.

Looking at the source, `FromStr` prohibits the usage of newlines in the header value.

Which one is "correct"?

### Expected behavior

An API which communicates the constraints `HeaderValue` puts on its inner string more clearly.

### Server and client version

The issue is purely in the rust crate's API

Jarema:
This is an inconsistency I will address - caused by adding a validation to headers and subjects.

PR #1612, opened 2026-08-16
Return errors instead of panicking on malformed MSG/HMSG/-ERR input

lazureykis:
Four malformed inputs panic `Connection::try_read_op` instead of returning an `io::Error`, killing the `ConnectionHandler` task — the failure mode #1572 fixed for non-UTF-8 subjects. All four reproduce in debug and release.

**A lone `\r` in a header value is reachable from any publisher.** `str::lines` strips a trailing `\r` but not an interior one, so `a\rb` reaches `From<String> for HeaderValue` and trips its assert. The server forwards header blocks verbatim, so a raw `HPUB` from any client with publish permission on a subject you subscribe to is enough. Verified against nats-server 2.14.5: the subscriber's stream ends and the next `publish` returns `ErrorKind::Send`.

The parse path now uses a `pub(crate) HeaderValue::from_string` — fallible, and unlike `FromStr` it doesn't copy the value a second time per header per message. Kept crate-internal to avoid doing #1548's API design here; `TryFrom` isn't available while `From<String>` exists.

**`MSG`/`HMSG` lengths near `usize::MAX`** overflow `len + payload_len + 4`. Debug traps; release wraps past the bounds check and panics in bytes with `split_to out of bounds`. Now `checked_add`. The `total_len < header_len` check doesn't cover it.

**A bare `-ERR\r\n`** makes `&read_buf[5..len]` start past its own end.

Five regression tests in the `read_op` module, including one pinning that folded header values still join.

Ran locally: `cargo +nightly fmt --check`, clippy with the `ring` backend, the full `--features slow_tests,websockets` suite, `cargo +1.88.0 check`. Not `--all-features` — `fips` needs cmake, which I don't have; CI covers it.


async-nats/src/connection.rs
```diff
@@ -391,7 +413,11 @@ impl Connection {
                 }
                 value.truncate(value.trim_end().len());
 
-                headers.append(name, value.into_header_value());
+                // `lines` strips a trailing `\r`, but not an interior one.
+                let value = HeaderValue::from_string(value)
```
stasimus, line 417: This turns one bad header value into a teardown of the whole connection. `poll_read_op` maps the `InvalidInput` to `ExitReason::Disconnected`, so every subscription and pending request on the client gets flapped. It's worse with a JetStream push consumer: the unacked HMSG is redelivered after reconnect and the client ends up in a disconnect/reconnect loop until the message is gone. The server relays header blocks verbatim, so any client with publish permission can persist a value with an interior lone `\r`.

Still a clear improvement over the old panic, but did you consider rejecting just the message (or sanitizing the value) instead of failing the read?

stasimus:
Outside this diff, but same bug class as the one this PR fixes: `service/mod.rs:824` does `headers.insert(NATS_SERVICE_ERROR, err.status.as_str())`, which still goes through the panicking `From<&str>`. An endpoint handler that returns a multi-line status (anyhow chains are a common source) trips the assert in `header.rs` and kills the endpoint's serving task instead of sending the error back to the requester. Now that `HeaderValue::from_string` exists, that call site could migrate too, here or in a follow-up.

async-nats/src/connection.rs
```diff
@@ -212,9 +213,20 @@ impl Connection {
...
+                return Err(io::Error::new(
+                    io::ErrorKind::InvalidInput,
+                    "number of payload bytes after MSG overflows",
+                ));
+            };
+
             // Return early without advancing if there is not enough data read the entire
             // message
-            if len + payload_len + 4 > self.read_buf.remaining() {
+            if required > self.read_buf.remaining() {
```
stasimus, line 229: The `checked_add` catches literal wraparound but not absurd values just below it. A server sending `MSG a 1 18446744073709551000\r\n` passes the check, then `required > remaining` stays true forever and `poll_read_op` keeps appending to `read_buf` until OOM. Same story for the HMSG `total_len`. A sanity cap (`max_payload` from INFO, say) would turn this into `InvalidInput` like the overflow case.

stasimus:
The lone-CR hardening misses one spot: the `NATS/1.0 <status> <description>` line. `lines()` keeps an interior `\r` there too, and around `connection.rs:389` the description goes into `ServerOp::Message` without any `HeaderValue` validation. So `NATS/1.0 503 no\rresponders` sails through while the same bytes in a header value are now rejected.

async-nats/src/connection.rs
```diff
@@ -157,7 +157,8 @@ impl Connection {
         }
 
         if self.read_buf.starts_with(b"-ERR") {
-            let description = str::from_utf8(&self.read_buf[5..len])
+            // A bare `-ERR\r\n` has no description, so `len` can be below 5.
+            let description = str::from_utf8(self.read_buf.get(5..len).unwrap_or_default())
```
stasimus, line 161: Pre-existing, but since this line is being rewritten anyway: the hardcoded offset 5 assumes byte 4 is a space without checking it. `-ERRfoo bar\r\n` passes the `starts_with` and parses as `ServerOp::Error("oo bar")` instead of being rejected.

stasimus:
Small one in the same parsing loop: when a header line has no `:`, the error at `connection.rs:402` says "no header version line found" even though the version line was present and fine. Sends you debugging the wrong thing.

stasimus:
Nit next to the new guard: the message at `connection.rs:332` says the header bytes were "greater than or equal to" the total, but equality is accepted (`header_len == total_len` with an empty payload parses fine). The check is a strict less-than.

async-nats/src/header.rs
```diff
@@ -415,6 +415,13 @@ impl From<String> for HeaderValue {
 }
 
 impl HeaderValue {
+    /// Like `From<String>`, but reports invalid values instead of panicking, and unlike
+    /// `FromStr` it reuses the allocation.
+    pub(crate) fn from_string(inner: String) -> Result<Self, ParseHeaderValueError> {
```
stasimus, line 420: The validate-then-wrap logic now lives in three places: here, `FromStr`, and `From<String>`. `FromStr` could be `Self::from_string(s.to_string())` and `From<String>` could delegate with an expect, so a future change to the value rules can't miss one of them and quietly diverge wire parsing from the public API.

async-nats/src/connection.rs
```diff
@@ -322,7 +334,17 @@ impl Connection {
                 ));
             }
 
-            if len + total_len + 4 > self.read_buf.remaining() {
+            let Some(required) = len
```
stasimus, line 337: This block is a copy of the MSG one above with a different error string. A small shared helper for the required-length math (or an `.ok_or_else(...)?` one-liner) would keep the two paths from drifting. The `len` shadowing inside the `and_then` closure also reads a bit oddly.

PR #1620, opened 2026-08-27
fix: FetchBuilder::priority overwrote batch size instead of setting priority

Jarema:
`FetchBuilder::priority(n)` (behind `server_2_12`) did `self.batch = priority` — the struct had no priority field, so on the fetch path the priority was silently dropped (`messages()` sent `priority: None`) and the batch size was clobbered to the priority value. The `stream()` path was unaffected.

Changes:
- Add `priority: Option<usize>` to `FetchBuilder` (feature-gated like the rest), set it in `priority()`, and pass it through in `messages()` — matching `StreamBuilder`.
- New test `prioritized_fetch`: prioritized consumer, subscribes on the `CONSUMER.MSG.NEXT` API subject to inspect the actual pull request, asserts `batch: 10` and `priority: 5`, then receives all 10 messages.

Verified: test fails before the fix (request carried `batch: 5`, no priority), passes after; `prioritized_pull_consumer` and the other fetch tests stay green; `cargo +nightly fmt` and `cargo clippy --all-features --tests` clean.

https://claude.ai/code/session_01XbNCnGpbvHuv1SQwDBC6nf


AdamPayzant (approved):
LGTM

bruth (approved):

PR #1621, opened 2026-09-01
fix(object_store): reading an object no longer panics on a JetStream timeout

macrauder:
## The bug

`Object`'s `AsyncRead` impl creates its ordered consumer lazily on the first
poll and `unwrap()`s the result:

```rust
// jetstream/object_store/mod.rs
context.get_consumer_from_stream(...).await.unwrap().messages().await
```

A JetStream request that ages past the client's timeout returns
`ConsumerErrorKind::TimedOut` here, so reading an object **panics on a runtime
worker thread** instead of returning an error. The caller never gets an `Err`
it could retry — it gets a poisoned task, and in a binary built with
`panic = "abort"` it gets a dead process, taking every unrelated in-flight
request on that runtime with it.

We hit this in production twice within one minute: two front-door processes
aborted (SIGABRT) mid-fetch, each killing connections belonging to other
callers. The object being read was intact and the request that produced it had
already succeeded — a slow JetStream round-trip was all it took.

## The fix

Both unwraps on that path become errors. Neither changes a public signature:

- `create_consumer(...)` → `map_err` into `StreamError`, preserving the
  timeout classification (`ConsumerErrorKind::TimedOut` →
  `StreamErrorKind::TimedOut`). The future's declared output is already
  `Result<Ordered, StreamError>`.
- `subscription.unwrap()` in `poll_read` → `map_err` into `io::Error`, which
  is what the surrounding code already returns for subscription failures a few
  lines below.

Removing the panic leaves two further properties to keep, both of which the
first revision of this PR got wrong (thanks to the review bot for catching
them):

- **The finished future must not stay on the `Object`.** With the `unwrap()`
  gone, `poll_read` returns an error and the caller may well retry the read —
  and the completed `async` block sitting in `subscription_future` would then
  be polled again, which panics with "`async fn` resumed after completion".
  That is the same process abort, one step further along. The future is now
  taken out of the `Object` before it is polled and put back only while it is
  still `Pending`, so a retry builds a fresh one and asks the server again.
- **The classification has to survive the `io::Error` boundary**, which is all
  an `AsyncRead` caller ever sees. Mapping every subscribe failure to
  `ErrorKind::Other` would have thrown away the `TimedOut` the line above works
  to preserve. A timeout now arrives as `io::ErrorKind::TimedOut`, with the
  `StreamError` retained as its source.

## Tests

Two integration tests against a real server, in `tests/object_store.rs`. Both
were confirmed failing before the fix:

| test | without the fix |
|---|---|
| `read_after_a_failed_subscribe_is_an_error_not_a_panic` | panics: "`async fn` resumed after completion" at `object_store/mod.rs` |
| `a_timed_out_read_keeps_its_kind` | `assertion left == right failed: left: Other, right: TimedOut` |

The first makes the subscribe fail deterministically by removing the bucket
between `get` and the first read, then reads twice. The second drops the server
after `get`, so the consumer request has nobody left to answer it.

`cargo test --test object_store` is green (17 tests), as are
`cargo +nightly fmt -- --check` and `cargo clippy --tests --examples --benches`.

## Notes

Cut from the `async-nats/v0.49.1` tag. Happy to rebase onto `main` — say the
word.


macrauder:
Pushed an update. The automated review on the first revision flagged two
follow-ups and both were real, so the PR now covers them.

**The finished future must not stay on the `Object`.** With the `unwrap()`
gone, `poll_read` returns an error and a caller may retry — and the retry
polled the already-completed `async` block still sitting in
`subscription_future`, which panics with "`async fn` resumed after
completion". That is the same process abort this PR exists to remove, one step
further along. The future is now taken out of the `Object` before it is polled
and put back only while it is still `Pending`, so a retry builds a fresh one.

**The classification has to survive the `io::Error` boundary**, which is all an
`AsyncRead` caller ever sees. Mapping every subscribe failure to
`ErrorKind::Other` threw away the `TimedOut` the line above works to preserve.
A timeout now arrives as `io::ErrorKind::TimedOut` with the `StreamError`
retained as its source.

Both have an integration test in `tests/object_store.rs`, and both were
confirmed failing before the fix:

| test | without the fix |
|---|---|
| `read_after_a_failed_subscribe_is_an_error_not_a_panic` | panics: "`async fn` resumed after completion" |
| `a_timed_out_read_keeps_its_kind` | `assertion left == right failed: left: Other, right: TimedOut` |

`cargo test --test object_store` is 17/17, `cargo +nightly fmt -- --check` and
`cargo clippy --tests --examples --benches` are clean, and each new test ran
10× without a flake.

One ask: CI has never run on this PR — no check runs on any head commit — so
`--all-features` clippy, the MSRV build and spellcheck have no result here. If
a maintainer can release the workflows I'll fix whatever they turn up, and I'm
happy to rebase onto `main` if that helps.

async-nats/src/jetstream/object_store/mod.rs
```diff
@@ -982,31 +982,58 @@ impl tokio::io::AsyncRead for Object {
 
         if self.has_pending_messages {
             if self.subscription.is_none() {
-                let future = match self.subscription_future.as_mut() {
+                // Taken out of `self`, and put back only while it is still
```
Jarema: please remove this comment.

